PR #29, opened 2019-07-18
Velociraptor/add vql updates

mgreen27:
Add update to Sysinternals query to cover all user hives.


scudette:
There seems to be something wrong? Is there an extra file?

Is it possible to write the artifact in terms of the
Windows.Registry.NTUser artifact? This way we can reuse and hide all
artifacts the need hkey_user in terms of the Windows.Registry.NTUser
artifact.

mgreen27:
HI Mike,

I chose to implement a lookup using Artifact.Windows.Sys.Users() instead of just globbing the c:\Users folder location because I am interested in pulling username. The relying on username from folder path is not ideal as in a domain you will often see c:\Users\username.domain paths. I liked the idea of using a standard template for enrichment but definitely open to a better way.

Seems like another file snuck in there (sublime top line save looks like).I can remove that.

Matt

mgreen27:
deleted the file and realised what you mean. Should be look at adding the Artifact.Windows.Sys.Users() lookup to the Windows.Registry.NTUser
artifact too? This way it will be a template way to look up user metadata in User hive queries.

scudette:
Yeah I totally agree - we need to find the different ntuser.dat from the user listing in the way you are doing it. Some of the ntuser.dat files that belong to service accounts are not found in the c:\users directory.

I just checked and realized that Windows.Registry.NTUser does use a simple glob which is not good enough. I was actually thinking of  Windows.Registry.NTUser.Upload which does the right thing.

So I think we need to refactor the artifacts a bit:
-  Windows.Registry.NTUser should be made to glob from all hives (We also need to use expand() to get the full path from the home directory path).
- All other artifacts that want to use hkey_users should be made to use  Windows.Registry.NTUser.

mgreen27:
Closing this PR and will reopen for the discussed changes.

PR #36, opened 2019-07-31
Velociraptor/add vql mods

mgreen27:
First go at adding the NTUser changes discussed here - https://github.com/Velocidex/velociraptor/pull/29
Output should be the same as the old version with additional SYSTEM hives and feilds enabled by leveraging Artifact.Windows.Sys.Users(). The additional fields are UserName and other user metadata that is useful during a query to pull out at the same time.

I also added a commit on a slight fix on some redundant VQL.


mgreen27:
image

image

PR #48, opened 2019-08-17
velociraptor/Add improvement to Binary Fetch

mgreen27:
I added some slight mods to the binary fetch utility after testing.

Added binPath to allow configuration of specific path (default is still %TEMP%). For example adding binPath = c:\Program Files\Velociraptor\bin works nicely :)
I also modded "\\" outside the expand function to accommodate this.

Potential things for us to look at in future:
1) built in copy /  move / delete functions to remove the need to shell out which can be noisy for process monitoring
2) reporting of failure - if hash doesn't match nothing passed to calling artefact.

### DIFF
--- a/artifacts/definitions/Windows/Utils/FetchBinary.yaml
+++ b/artifacts/definitions/Windows/Utils/FetchBinary.yaml
@@ -17,8 +17,7 @@ parameters:
 sources:
   - queries:
       - |
-        LET download = SELECT FullPath as TempPath,
-            expand(path=binPath) + "\\" + basename(path=binaryURL) as FullPath,
+        LET download = expand(path=binPath) + "\\" + basename(path=binaryURL) as FullPath,
             hash(path=Content) as Hash, 
             {
                 SELECT * FROM execve(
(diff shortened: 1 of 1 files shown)


artifacts/definitions/Windows/Utils/FetchBinary.yaml
```diff
@@ -17,8 +17,7 @@ parameters:
 sources:
   - queries:
       - |
-        LET download = SELECT FullPath as TempPath,
-            expand(path=binPath) + "\\" + basename(path=binaryURL) as FullPath,
+        LET download = expand(path=binPath) + "\\" + basename(path=binaryURL) as FullPath,
```
scudette: I dont think this is valid VQL - LET must be followed by select

mgreen27: it worked for me testing via the gui in 0.32 release.
image

scudette:
I reworked FetchBinary to be more general purpose - take a look at the new version and see if this PR is still relevant.

mgreen27:
looks much better

PR #100, opened 2019-10-02
Mgreen27.velociraptor.git/add kerbroasting

mgreen27:
Added HUNT report for Kerbroasting content.

I also added a monitoring artefact using watch_evtx().


scudette:
This seems broken - maybe need to merge?

PR #217, opened 2020-01-18
Add PermanentWMIEvents multiple namespaces

mgreen27:
Add check of root/default and root/subscription namespaces by default.
Add namespace column into output


Uchedanieln:
- [ ] Optional

PR #284, opened 2020-03-23
Update Services.yaml

sjosz:
Included the option to calculate hash and to return certificate information

### DIFF
--- a/artifacts/definitions/Windows/System/Services.yaml
+++ b/artifacts/definitions/Windows/System/Services.yaml
@@ -1,4 +1,4 @@
-name: Windows.System.Services
+name: Custom.Windows.System.Services
 description: |
   List all the installed services.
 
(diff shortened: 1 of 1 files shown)


artifacts/definitions/Windows/System/Services.yaml
```diff
@@ -5,21 +5,37 @@ description: |
...
+                 SELECT timestamp(epoch=Mtime.Sec) as Created
+                 FROM stat(filename=servicesKeyGlob + Name, accessor='reg')
                } AS Created,
                {
                  SELECT ServiceDll FROM read_reg_key(globs=servicesKeyGlob + Name + "\\Parameters")
-               } AS ServiceDll
+               } AS ServiceDll,
+               parse_string_with_regex(regex=
+                 ['^\\"(?P<AbsolutePath>[^"]+)','(?P<AbsolutePath>^[^"]+)[ ][-~/]+','(?P<AbsolutePath>^[^ ]+$)'], 
+                 string=PathName) as Parsed
```
scudette: Since there is really only one field here this might look a bit better:

```
          parse_string_with_regex(regex=
                 ['^\\"(?P<AbsolutePath>[^"]+)','(?P<AbsolutePath>^[^"]+)[ ][-~/]+','(?P<AbsolutePath>^[^ ]+$)'], 
                 string=PathName).AbsolutePath as AbsolutePath
```

2. I think the `\\` might not be necessary ` ['^"(?P<AbsolutePath>[^"]+)', `

3. I dont know if it makes sense to require `[-~/]+` after the command - this will not match a lot of commands (like `sc create` or `net user`) maybe drop it?

sjosz: 1. Great, that looks better indeed, I changed it.

2. The `\\` is sometimes necessary since the value PathName can start with `\\` : 

image

3. I can remove the `~` since this would not occur very often, but the rest are required to extract the absolute path from the following services entries in registry:

`C:\\Windows\\System32\\svchost.exe` -k wsappx
`C:\\Windows\\system32\\msiexec.exe` /V

scudette: 2. I dont think it actually starts with `\` - it is just showing you that because it starts with `"` and it is inside a string so it needs to be quoted. So the case you showed above starts with `"` actually.

3. If you only captured until the space it would still work and would get commands with args that do not contain flags (like `sc create`). 
```
'(?P<AbsolutePath>^[^"]+) '
```

sjosz: 2. Check, I removed it and it works!

3. Also works, thanks.

scudette:
Thanks!

PR #929, opened 2021-02-26
Fix 32bit module enumeration

clayscode:
This PR makes the modules plugin work on 32bit versions of Windows and fixes a potential bug with enumerating 32bit modules on 64bit Windows.


clayscode:
Hmm, thinking about it, instead of a structs.go file, I guess gen.go could be modified to produce both a 32 bit and 64 bit version of win32_windows.go?

PR #957, opened 2021-03-08
Fix SFTP Upload with write-only permissions

davidjmaria:
Made change to catch ErrSSHFxPermissionDenied error when uploading files using an account without read permissions. Whenever an upload is attempted when read permissions are not available, the ErrSSHFxPermissionDenied error will be caught and the user will be warned that the size of the uploaded file cannot be verified.

I have manually tested SFTP uploads with the resulting binaries using an AWS managed services SFTP server, as well as on a traditional OpenSSH SFTP server.
Fixes: #950

### DIFF
--- a/vql/tools/sftp_upload.go
+++ b/vql/tools/sftp_upload.go
@@ -203,7 +203,13 @@ func upload_SFTP(ctx context.Context, scope vfilter.Scope,
 
 	file.Close()
 	check, err := client.Lstat(fpath)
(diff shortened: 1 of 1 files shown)


davidjmaria:
Unsure why the check failed, it was successful on my fork:
https://github.com/davidjmaria/velociraptor/pull/2/checks?check_run_id=2061354694
It it be possible to rerun it?

PR #964, opened 2021-03-11
PR to add unzip plugin to velociraptor 

clayscode:
This plugin adds the ability to unzip a file (e.g. `SELECT * FROM unzip(filename="SomeFile.zip", output_directory="NewDirectory)` ). If you could, please review line 58 to suggest a better alternative since just appending a glob to the base file feels a little hacky.


clayscode:
Wow, really screwed this one up. Didn't realize I screwed up my rebase so badly

PR #1061, opened 2021-05-08
Add Cortex Analyzer artifact

weslambert:
This PR adds the ability to submit jobs to TheHiveProject's Cortex observable analysis engine to gather additional context around data obtained by Velociraptor (such as a hash, domain, or IP). 

Currently, the analyzer will submit a job to all enabled analyzers that support the provided data type.  A report is returned for each enabled/supported analyzer upon artifact completion.

More information about Cortex can be found here: 
https://github.com/TheHive-Project/Cortex
Information about Cortex analyzers can be found here: 
https://github.com/TheHive-Project/CortexDocs/blob/master/analyzer_requirements.md#introduction

### DIFF
--- a/artifacts/definitions/Server/Enrichment/CortexAnalyzer.yaml
+++ b/artifacts/definitions/Server/Enrichment/CortexAnalyzer.yaml
@@ -1,0 +1,98 @@
+name: Server.Enrichment.CortexAnalyzer
+description: |
+
+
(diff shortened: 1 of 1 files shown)


scudette (approved):
Thanks! Do you think you could add a test? For a test we will need the json output of the http_client from the hive. I am happy to do the test in a new PR but will need samples so we can mock out the http_client.

PR #1140, opened 2021-07-13
Add Windows.Detection.Yara.Process

mgreen27:
Consolidated a few yara detection rules into a standard Windows.Detection.Yara.Process confirming to the ntfs and glob yara standard.

Windows.Detection.ProcessMemory.CobaltStrike
Windows.Detection.RemoteYara.Process
Windows.Detection.ProcessMemory
= Windows.Detection.Yara.Process


mgreen27:
Searching for CobaltStrike will still pop Windows.Detection.Yara.Process and the CobaltStrike yara rule is default for this artifact.

image

artifacts/definitions/Windows/Detection/Yara/memory.yaml
```diff
@@ -1,26 +1,34 @@
...
-  - https://malpedia.caad.fkie.fraunhofer.de/details/win.cobalt_strike
-
+  Regex parameters can be applied for process name and pid for targeting. The 
+  artifact also has an option to upload any process with Yara hits.  
+  
+  Note: the Yara scan will stop after one hit.
+  
+type: CLIENT
 parameters:
   - name: ProcessRegex
```
scudette, line 20: we seem to have lost the whitelist from above? this could be useful

mgreen27: I removed it as I never really used it and post processed whitelists. Do you think its worth adding back in?

scudette: I think it is useful - sometimes we need to avoid scanning sensitive processes (like lssas.exe) because it will also trigger EDRs as well.

mgreen27: just added whitelist capability in with placeholder path.

PR #1414, opened 2021-12-07
update for clarity

randomaccess3:


artifacts/definitions/Windows/Registry/AppCompatCache.yaml
```diff
@@ -16,6 +16,6 @@ sources:
               SELECT Filename, Data FROM read_file(
                   filenames=AppCompatKeys.FullPath, accessor='reg')
           }, query={
-              SELECT Filename AS Key, name as Name, epoch, time
+              SELECT Filename AS Key, name as Name, epoch, ModificationTime
```
mgreen27, line 19: I think you will need to make this: time AS ModificationTime

randomaccess3: Yep, my bad, will fix

PR #1427, opened 2021-12-13
Update Info.yaml

Assem-Morad:
I have enriched the artifact with the 'HostIP' column via the "Win32_NetworkAdapterConfiguration" WMI class.

### DIFF
--- a/artifacts/definitions/Generic/Client/Info.yaml
+++ b/artifacts/definitions/Generic/Client/Info.yaml
@@ -26,7 +26,13 @@ sources:
                      then={
                          SELECT Domain FROM wmi(query='SELECT Domain FROM win32_computersystem')
                      })
-               } AS ADDomain
+               } AS ADDomain,
+                {
+                  SELECT IPAddress[0] FROM wmi(
+                    query="SELECT * from Win32_NetworkAdapterConfiguration" )
+                        WHERE DefaultIPGateway =~ ''
+               } AS HostIP
+               
         FROM info()
 
   - name: Users


artifacts/definitions/Generic/Client/Info.yaml
```diff
@@ -26,7 +26,13 @@ sources:
                      then={
                          SELECT Domain FROM wmi(query='SELECT Domain FROM win32_computersystem')
                      })
-               } AS ADDomain
+               } AS ADDomain,
+                {
+                  SELECT IPAddress[0] FROM wmi(
+                    query="SELECT * from Win32_NetworkAdapterConfiguration" )
+                        WHERE DefaultIPGateway =~ ''
```
scudette, line 33: This is a no-op ? Any string matches the empty string regex?

Also do you think we should report all the ip addresses?  Many systems might have several interfaces and multiple ip addresses (definitely ipv4 and ipv6)

Assem-Murad: Regarding the empty string regex, my point is to match DefaultIPGateway that has value as the DefaultIPGateway with Null value isn't important in our case. 
and regarding the machines that have several interfaces, I have enhanced my query to collect all IP addresses into 1 list, moreover, this Query Supports both versions IPv4 and IPv6 and we can split it into 2 columns as shown below 
Gs

scudette: I am actually wondering if we should collect this additional information in a different artifact source? The BasicInformation source is used by the interrogation service to index some things (like hostname and labels) so i think maybe we dont want to change it too much - perhaps the best way is to add an additional os specific artifact source to collect these extra pieces like ip addresses.

Assem-Murad: Actually, I was thinking the same about creating a new artifact that has some details about network configuration as a new artifact.

PR #1462, opened 2022-01-05
WIP: dead disk analysis

lucebac:
Hello everyone,

I've been playing around with some ideas regarding dead disk analysis (#888 and #1425). So far I managed to make the `glob` command work on NTFS disk images to retrieve file meta data from there. I have not touched anything but file access yet, however, other accessors might already work by chance (untested, though). The implementation can be summed up roughly as follows:

* I replaced the implementation of the standard `file` accessor with a new `PassthruAccessor` (was `OSFileSystemAccessor` before). This allows to seamlessly switch the backing accessor which ultimately accesses the data. Technically, this even allows to use any available accessor as the backing "data source" and not only the `ntfs` one which I have been testing with. If there is no `--device` specified, the passthru falls back to the OS-specific file accessors.
* I modified the NTFS accessor and extracted OS-specific code into different files. Hence, the NTFS accessor now also works on non-Windows OSes (most of the code already was cross-platform, just the VSS stuff and volume discovery did not work because of WMI).

An exemplary command line for running a glob on a disk image would be: `./output/velociraptor-v0.6.3-rc1-linux-amd64 --device /dev/nbd1p2 --device_accessor ntfs query 'SELECT * FROM glob(globs="\\windows\\*")'`. You may need to adjust the device path, though.

Please note: this is by no means ready for an actual review. The code is more of a POC than anything else (e.g. license headers are missing all over the place, variable and function names may be inconsistent or non-descriptive, etc). I just wanted to share this very early state to kindly ask for some feedback on what I have been doing and if this might be the way to go for a full-fledged implementation.


vql/windows/filesystems/ntfs_nonwindows.go
```diff
@@ -0,0 +1,30 @@
+//go:build !windows
+// +build !windows
+
+package filesystems
```
scudette, line 4: This is great to make the accessor work on linux but it has to be done in a way that we can still do the same things on linux or windows - so allow a windows binary to also deal with images.

Ultimately this will be a specialised accessor that remaps a directory based on the remapping configuration.

lucebac:
Thanks for the reply! You propose some great ideas which I had not yet thought of. The remapping config is much more versatile than what I had in mind - most importantly this allows to have more than one source of data which comes in handy if one wants to evaluate several disk images using only one Velociraptor instance (e.g., when the analysis target uses more than one partition for data storage).

What do you think about merging `file_mappings` and `ntfs_mappings` into something more generic and just specify the type of the source, maybe like this:
```yaml
Client:
   remappings:
      - source: /dev/sdc3
        map_at: /diskA
        type: ntfs
      - source: /mnt
        map_at: /diskB
        type: file
      - source: /data/backup.zip
        map_at: /diskC
        type: zip
```
Maybe we could then also use something similar to the `\\.\<device>\...` structure internally for the mappings and use `map_as: <device>` instead of `map_at: <path>`. IIRC the NTFS parser already maps devices this way anyway so it might be useful to move this logic into Velociraptor and leave just and only the parsing to the go-ntfs package.

> The end result is to get the **same** artifacts VQL that we use normally to run as well as possible inside the disk image environment. Of course artifacts that use things like WMI, pslist etc will be meaningless so we do need to disable them (so they return no results) but those that use glob, parsing files etc should work fine.

Yes, this is a requirement in my opinion as well; this is also why I refrained from touching any artifact and only modified the go code to create my current implementation. I think re-registering or un-registering certain accessors/plugins depending on if we are in a mapping environment is the way to go here because this implicitly also disables the artifacts which would not work in the disk image anyway. Yet, this probably needs some additional work because I am afraid that artifacts currently just fail instead of returning an empty result when plugins are missing - however, I don't know that for sure and need to verify this.

> It might be a bit tricky at first to come up with the proper mapping config but we can make this process a lot easier by e.g. opening the /etc/fstab file and building maps based on that (or look inside the disk image and just assume one containing Windows is C:), or just let the user write the config by hand (some of this is very hard to automate).

I think automatically creating the mapping config should be postponed to when there is a working implementation of the mapping. And until then, the mappings must be configured by hand because, as you stated, generating these mappings from mounted/attached external storage devices will certainly be a hard task.

scudette:
Yes we will probably need to re-register plugins as well for noop plugins in remapping mode so the client behaves as if it is simulating the target OS. For example when running a linux client which normally does not have the wmi plugin, but the remapping config tells it to behave like a windows machine we need to register a noop wmi plugin 

Otherwise we will get a warning no wmi plugin found - or maybe this is fine? 

Eitherway it is probably not a huge priority - i think getting the file remapping working should be the first thing - implement a second file accessor with a relative root - or maybe we can modify the current file accessor to accept a relative root (by default will be /)

scudette:
Yes I think the remapping config suggestion is good. This can actually be useful in the case of monitoring containers, VMs etc from the host point of view - we might be able to use the same method to make the client appear like it is running inside the container.

lucebac:
I just pushed 7549192 which implements how plugins could be replaced with a noop plugin if they are not applicable for a given configuration (e.g. WMI plugin on Linux host). I will work on a draft implementation for the remapping config next.

lucebac:
I got the remapping to work partially. It is currently possible to remap ntfs, zip, and file data sources and search them all as if they were different disks on the same host. 

However, there is a something odd with regards to the filters which match a path from a source against the list of provided `globs`. I will look into this when I'm back on monday and to figure out why this causes issues. 

This is the example config I am working with:
```yaml
analysis_target: windows
remappings:
  - source: /dev/nbd1p2
    map_as: "c:"
    type: ntfs
  - source: /tmp/velo.zip
    map_as: "d:"
    type: zip
  - source: /tmp
    map_as: "e:"
    type: file
```
All remapped sources contain the folder structure which would lead the following command to successfully retrieve the evtx files from the respective source: 
```bash
DEBUG=1 go run ./bin/ -v -c config.yml query 'SELECT * FROM collect(artifacts="Generic.Collectors.File", args=dict(`Generic.Collectors.File`=dict(Root="", Accessor="", collectionSpec="Glob\nWindows\\system32\\winevt\\logs\\Windows PowerShell.evtx\nWindows\\system32\\winevt\\logs\\System.evtx")))'
```

vql/filesystem/zip.go
```diff
@@ -358,9 +358,10 @@ func (self *ZipFileCache) Close() {
 // root scope. We keep a list of most recently used cached of zip
 // files for quick access.
 type ZipFileSystemAccessor struct {
-	mu       sync.Mutex
-	fd_cache map[string]*ZipFileCache
-	scope    vfilter.Scope
+	mu         sync.Mutex
+	fd_cache   map[string]*ZipFileCache
+	scope      vfilter.Scope
+	sourceFile string
```
scudette, line 364: Can you please explain why we need this?  Since the zip accessor usually calls a delegate it probably does not need to specially handle the remapping?

vql/filesystem/filesystem.go
```diff
@@ -97,46 +93,88 @@ func (self GlobPlugin) Call(
 			}
 		}
 
-		globber := glob.NewGlobber().WithOptions(options)
+		organizedGlobs := make(map[string][]string)
+		var globalGlobs []string
 
-		// If root is not specified we try to find a common
-		// root from the globs.
+		// If root is not specified we organize the globs by their root and evaluate
```
scudette, line 99: Can you explain why we need to refactor this code? The globber is independent from the accessors and i dont think will be affected by remapping.

bin/analysis_target_nonwindows.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+func applyAnalysisTarget(config_obj *config_proto.Config) error {
+	if config_obj.AnalysisTarget == "" {
+		return nil
+	}
+
+	switch config_obj.AnalysisTarget {
+	case "windows":
+		vql_subsystem.MakeNoopPlugin("wmi")
```
scudette, line 18: We obviously need to remove more than this plugin.

I am just wondering how do we keep these in sync as new plugins are introduced? This is why I was suggesting using the ACL system instead because each plugin should declare the required ACLs and so it will naturally work that running the plugin will receive permission denied when running in this mode.

But probably this will be needed anyway to provide "stub" plugins for different OS platforms - otherwise we will receive symbol not found errors on Linux for wmi for example.

lucebac: Yes, this definitely needs to be done to more than the wmi plugin. I picked the wmi plugin only because I knew it did not work on linux to see how my idea of replacing plugins feels like. And to be honest: although it works, I do not really like the conditional compilation either. But I did not know of a more appropriate method of doing certain things based on the host OS, so I went for the "ugly" way. However, I will have a thorough look into the ACL system and see how it works out if we use ACLs instead of conditional compilation.

vql/filesystem/zip.go
lucebac: I will revisit what I did here. Seeing my changes right now, it appears to me that I had something in mind regarding the pathspec which was conceptually wrong.

vql/filesystem/filesystem.go
lucebac: In general, I agree with you: the remapping does not affect the globber and how it uses the accessor(s). The refactoring was mostly needed because of the fact that the upstream code uses only one single accessor for all globs. While this works perfectly fine when there is only the host's filesystem to consider, it trips when the remapping is in place: depending on the path which is to be accessed, different accessors need to be used because of the remapping (or even all of them if a wildcard device or no specific device is selected). Therefore, the changeset introduces some code which organizes the globs by their device/root before the underlying data is accessed in the inline function.

My implementation is most probably not the best solution to this but it works to some degree and may serve as a starting point to build on - if we decide to dump it, be it so. In that case we will at least know that this is not the way the remapping may be implemented.

lucebac:
While working on your suggestions from the weekend I had an idea: would you consider it beneficial if all FileSystemAccessors receive a `PathSpec` instead of the filename as a string? I suspect this breaks a lot of the code but as it seems you want want to replace the/all url-based paths with pathspecs in the future this might be a fit.

scudette:
Yes all accessors do take a pathspec if they need it - it wont break any code because we can do it in a backwards compatible way. A pathspec is just a way to pass structured information to an accessor than a flat "filename" which may not have enough information.

Currently things like zip accessor use pathspec to figure out where the zip is and which member to read out of it. I am not sure about all accessors though - which one were you thinking could benefit?

lucebac:
I specifically needed to add the pathspec to the ntfs accessor. Otherwise I had no chance to let it know which file it had to parse. Doing so involved adding `glob.PathSpecFromString` to every function of the `FileSystemAccessor` interface. This resulted in having "many" unneeded (de)serializations as the new workflow now is `create PathSpec -> ToString -> call interface -> PathSpecFromString -> use PathSpec`. This made me come up with the idea of just passing the PathSpec instead of doing all the conversions.

scudette:
I don't think this is needed for the functionality discussed in this pr. The file name provided to the accessor is given by the vql code which is supposed to be completely unaware of any mapping.

If it needed to include the mapping in the vql then we could not use the same vql in all contexts and that would defeat the purpose of making it transparent.

glob/accessor_common.go
```diff
@@ -171,21 +172,32 @@ func (self OSFileSystemAccessor) New(scope vfilter.Scope) (FileSystemAccessor, e
...
-func (self OSFileSystemAccessor) ReadDir(path string) ([]FileInfo, error) {
-	path = self.GetPath(path)
-	lstat, err := os.Lstat(path)
+func (self OSFileSystemAccessor) ReadDir(dir string) ([]FileInfo, error) {
+	pathSpec, err := PathSpecFromString(dir)
+	if err != nil {
+		return nil, err
+	}
+
+	fullpath := path.Join(pathSpec.DelegatePath, pathSpec.Path)
```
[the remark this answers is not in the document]
lucebac: > So you want to get the OSFileSystemAccessor to take a delegate pathspec so it can be rooted at any arbitrary accessor/path. This is a great idea but I think it is actually a completely different type of accessor - say we call it a MountedAccessor

I had not yet thought of a dedicated accessor for that but yes, this is essentially what I had in mind when I was playing around with different approaches for the implementation. Thinking about it, you may have just proposed a very good solution to the whole remapping thing because internally the `mount` accessor could just replace the `file` accessor if remappings are enabled and everything works out of the box.

Thinking two steps ahead, using pathspecs this way could perhaps also allow pathspec chaining (think of something like "read a file from a zip-compressed disk image"). I'm not sure if this would ever be needed, though, but even if, this would be something for the (far) future and imo there is other things which need to be set up first before things like chaining would make any sense.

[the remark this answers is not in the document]
scudette: No this is exactly how it works :-). Pathspec chaining is already used to recursively search zip files - see eg
https://docs.velociraptor.app/exchange/artifacts/pages/log4jvulnhunter/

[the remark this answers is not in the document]
lucebac: Isn't that only implement for zips by now? I thought of chaining in a more general way which is afaict not yet possible because the other accessors cannot handle pathspecs for now.

lucebac:
I've come across a structural problem for which I did not yet have an idea on how to solve this.

If I just re-register the file accessor with, say, a `RemappingAccessor` to redirect the FileSystemAccessor interface into another accessor which ultimately accesses the data, I end up with endless loops. Given the remapping
``` 
remapping:
  - source: /dev/nbd1p2
    type: ntfs
    map_as: /
```
the overwritten `file` accessor uses the `ntfs` accessor to read the data. However, the `ntfs` accessor in turn needs a `file` accessor to open the disk image. But the latter hands over everything to the `ntfs` accessor which results in having an endless chain of `file -> ntfs -> file -> ntfs -> ...` calls. This also applies to any other accessor which uses the `file` accessor to read their data.

I am tempted to go back to my initial approach but this would imply rewriting the glob plugin partially. If you have any other ideas, I would be glad to hear about them - maybe I am just overlooking something here.

scudette:
This is a good point - the remapping accessor needs to capture an instance of the other accessors (i.e. not use the resolver to get the accessor based on path at all) - it should already have the delegate object on hand during creation time and through the life of the program.

So call glob.GetAccessor() during instantiation and hold onto the accessors in a map keyed by "mount point". Then re-register the accessors and at runtime just use the correct accessors from that map without calling glob.GetAccessor()

This avoids the loop.

lucebac:
I am not sure if I'm getting what you suggest. The `ntfs` accessor only calls `glob.GetAccessor` to get the accessor for the ntfs disk image when actually reading from the underlying Reader (the PagedReader does that transparently):

https://github.com/Velocidex/velociraptor/blob/2ebfba0525a8838b5b67418f78d172b071033efb/vql/readers/paged.go#L151

I currently do not see how I can avoid the issue if I'm replacing the `file` accessor even if I keep a reference to the original `ntfs` accessor. During runtime the `ntfs` accessor still only sees the replaced `file` accessor instead of the original one. This was an implicit advantage of the rewritten `Call` function of the glob plugin: I was then able to retrieve and use the accessor of the remapping instead of the one returned by `glob.GetAccessor`.

scudette:
Would it help if you can keep a reference to the old (unmodified) file accessor? maybe register it as "original_file" or something?

lucebac:
I don't think so. Because even if I kept an `original_file` reference I would doubt it is possible to use it where we would need it: it is, at least currently, just not possible to correctly solve situations like "the `ntfs` accessor requests the `file` accessor but I need to return `original_file` because I replaced it". I unfortunately do not know how to find out who requests an accessor so that I can return the appropriate `file` accessor.

scudette:
I will have some time to play with this next week hopefully.

lucebac:
Greatly appreciated! I will push my latest changes during the course of the day so that you have everything at hand I adjusted so far.

lucebac:
Apparently, being off the project for a few days helps solving seemingly complex issues. I did not manage to implement the remapping without at all touching the glob implementation, however, the current approach only modifies very few lines and everything else remains untouched (the commit diff may indicate otherwise but I had to revert the changes I introduced previously; only the `deviceManager` block is new now).

scudette:
I'm going to close this one since this https://github.com/Velocidex/velociraptor/commit/60114e12652af1dc8f1f887415d1ca04e90e592a was submitted.

It pretty much does what is intended at this stage. We still need to think about controlling plugins for dead disk but this is a good start.

PR #1526, opened 2022-01-27
Update descriptions and keypath

randomaccess3:
Added description to describe that two fields are not being parsed but are also unreliable so it's ok.
Changed the description of userfilter to 'username' instead of 'user id' in case analysts confuse this for the user SID
Also included the KeyPath as a hidden field as it may be valuable to an investigator.


randomaccess3:
Making additional changes to previous artifact, close request for now

PR #1546, opened 2022-02-08
Remove C time and updating naming

randomaccess3:
the C time is not relevant in LNK file parsing for file access. The B time (file creation) is significantly more important and hasnt been included.
Removed all references to C time and renamed mtime and btime to SourceModified and SourceCreated respectively


mgreen27:
@randomaccess3 our test failed after you ran your update. These are designed to test for changes that impact content. Instead of me just fixing, if your interested we can catch up and I can run you through how to mod these too fir success?

randomaccess3:
Yeah can do. My guess is if there's a LNK file test then the LNK file test is incorrect.

mgreen27:
yeah exactly, its literally just changing the test to match your changes!

PR #1638, opened 2022-03-09
Add Generic.Detection.HashHunter and tests

mgreen27:
This artifact enables searching for hashes with best performance.
The artifact takes a glob targeting input, then generates a hash for each file in scope to compare to several types of hash lists provided by the user.

NOTE: I noticed there is also a hash optimisation PR, will aim to also update once included.

### DIFF
--- a/artifacts/definitions/Generic/Detection/HashHunter.yaml
+++ b/artifacts/definitions/Generic/Detection/HashHunter.yaml
@@ -1,0 +1,95 @@
+name: Generic.Detection.HashHunter
+author: "Matt Green - @mgreen27"
+description: |
+    This artifact enables searching for hashes.
+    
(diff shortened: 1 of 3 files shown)


eduardomcm:
@mgreen27 for large hash lists it might be better using a CSV. My team has been using this one instead https://github.com/eduardomcm/VelociraptorCompetition/blob/main/Windows.Detection.HashHunt.yaml

Could you add that option to this artifact?

mgreen27:
Do you think it's worth switching to hash based lookup?

The current takes a line or space delimited entry for tactical use but we can always look at csv tools fallback too.

eduardomcm:
After thinking about my use cases, I think I'll end up using your solution over CSVs Matt :) Thanks for the artifact!

PR #1809, opened 2022-05-25
artifactset: add ability to select named sources

jeffmahoney:
A comment on PR #1790 mentioned that there was a shortcoming vs using
CSV fields in that individual sources could not be selected.  This
commit iterates over the sources for each artifact and presents them
in the UI instead of just the bare artifact name.


jeffmahoney:
This PR works with the artifacts we have now and doesn't depend on any validation performed as the result of #1808.

PR #1906, opened 2022-06-25
Japanese translation

YamatoSecurity:


predictiple:
We might need to add special font support for this.

predictiple:
The Japanese option also needs to be added to the other non-English translation files.

PR #1927, opened 2022-07-07
Windows.Attack.IncorrectImagePath

AmgdGocha:
It is for creating an artifact that looks for processes with known names that are being loaded from unexpected locations.

### DIFF
--- a/artifacts/definitions/Windows/Attack/IncorrectImagePath.yaml
+++ b/artifacts/definitions/Windows/Attack/IncorrectImagePath.yaml
@@ -1,0 +1,127 @@
+name: Windows.Attack.IncorrectImagePath
+
+Description: |
+  Some malware are hiding in plain text by masqurading a legitimate executable name.
+
+  This artifact looks for processes with known names that are being loaded from unexcpected 
+  locations.
+
+reference: 
+  - https://www.sans.org/posters/hunt-evil/
+  - https://github.com/teoseller/osquery-attck/blob/master/windows-incorrect_path_process.conf
+
+author: Amged Wageh
+
+parameters:
+   - name: expected_
+     type: json_array
+     default: |
+        [
+          {
+            "ProcessName": "system",
+            "ImagePaths": []
+          },
+          {
+            "ProcessName": "csrss.exe",
+            "ImagePaths": [
+              "c:\\windows\\system32\\csrss.exe"
+            ]
+          },
+          {
+            "ProcessName": "smss.exe",
+            "ImagePaths": [
+              "c:\\windows\\system32\\smss.exe"
+            ]
+          },
+          {
+            "ProcessName": "services.exe",
+            "ImagePaths": [
+              "c:\\windows\\system32\\services.exe"
+            ]
+          },
+          {
+            "ProcessName": "wininit.exe",
+            "ImagePaths": [
+              "c:\\windows\\system32\\wininit.exe"
+            ]
+          },
+          {
+            "ProcessName": "svchost.exe",
+            "ImagePaths": [
+              "c:\\windows\\system32\\svchost.exe",
+              "c:\\windows\\syswow64\\svchost.exe"
+            ]
+          },
+          {
+            "ProcessName": "runtimebroker.exe",
+            "ImagePaths": [
+              "c:\\windows\\system32\\runtimebroker.exe"
+            ]
+          },
+          {
+            "ProcessName": "lsaiso.exe",
+            "ImagePaths": [
+              "c:\\windows\\system32\\lsaiso.exe"
+            ]
+          },
+          {
+            "ProcessName": "taskhostw.exe",
+            "ImagePaths": [
+              "c:\\windows\\system32\\taskhostw.exe"
+            ]
+          },
+          {
+            "ProcessName": "lsass.exe",
+            "ImagePaths": [
+              "c:\\windows\\system32\\lsass.exe"
+            ]
+          },
+          {
+            "ProcessName": "winlogon.exe",
+            "ImagePaths": [
+              "c:\\windows\\system32\\winlogon.exe"
+            ]
+          },
+          {
+            "ProcessName": "explorer.exe",
+            "ImagePaths": [
+              "c:\\windows\\explorer.exe",
+              "c:\\windows\\syswow64\\explorer.exe"
+            ]
+          },
+          {
+            "ProcessName": "conhost.exe",
+            "ImagePaths": [
+              "c:\\windows\\system32\\conhost.exe"
+            ]
+          },
+          {
+            "ProcessName": "dllhost.exe",
+            "ImagePaths": [
+              "c:\\windows\\system32\\dllhost.exe",
+              "c:\\windows\\syswow64\\dllhost.exe"
+            ]
+          },
+          {
+            "ProcessName": "wmiprvse.exe",
+            "ImagePaths": [
+              "c:\\windows\\system32\\wbem\\wmiprvse.exe",
+              "c:\\windows\\syswow64\\wbem\\wmiprvse.exe"
+            ]
+          }
+        ]
+
+sources:
+  - precondition:
+      SELECT OS From info() where OS = 'windows'
+
+    query: |
+      LET running_processes <= SELECT Name, Pid, Ppid, {SELECT Name FROM pslist(pid=Ppid)} AS ParentProcessName, Exe, CommandLine, Username, CreateTime FROM pslist()
+      
+      SELECT * FROM foreach(
+          row={SELECT ProcessName, ImagePaths FROM expected_paths},
+          query={
+              SELECT * FROM running_processes 
+              WHERE Exe!="" AND lowcase(string=Name)=ProcessName AND NOT lowcase(string=Exe) in ImagePaths
+          }
+      )


mgreen27:
Can you fix the formatting of your artifact?
image

AmgdGocha:
@mgreen27 I've
1- Fixed a letter capitalization
2- Added a `WHERE` clause to the `running_processes` to only triage the processes that exist in the CSV parameter
3- moved the function calls to the final `SELECT` statement.

artifacts/definitions/Windows/Attack/UnexpectedImagePath.yaml
```diff
@@ -0,0 +1,73 @@
...
+          row=expected_paths_lookup,
+          query={
+              SELECT * FROM running_processes 
+              WHERE ImagePath!="" AND lowcase(string=ProcessName)=ProcName AND NOT lowcase(string=ImagePath) IN ExpectedPaths
+          }
+      )
+
+      SELECT PID, ProcessName, ImagePath, CommandLine, Username, TokenIsElevated, StartTime, EndTime, 
+        PPID, process_tracker_callchain(id=PID)[-2].Data.Name As ParentProcessName, 
+        process_tracker_callchain(id=PID)[-2].Data.Exe As ParentImagePath,
```
scudette, line 65: Each call to the proecess tracker translates to a full pslist when the tracker has no implementation (i.e. it is not cached by default) so this will be very expensive and it is not necessary - you can just look up the same results for each field.

artifacts/definitions/Windows/Attack/UnexpectedImagePath.yaml
```diff
@@ -0,0 +1,73 @@
...
+
+      SELECT PID, ProcessName, ImagePath, CommandLine, Username, TokenIsElevated, StartTime, EndTime, 
+        PPID, process_tracker_callchain(id=PID)[-2].Data.Name As ParentProcessName, 
+        process_tracker_callchain(id=PID)[-2].Data.Exe As ParentImagePath, 
+        process_tracker_callchain(id=PID)[-2].Data.CommandLine As ParentCommandLine, 
+        process_tracker_callchain(id=PID)[-2].Data.Username As ParentUsername,
+        process_tracker_callchain(id=PID)[-2].Data.TokenIsElevated As ParentTokenIsElevated, 
+        process_tracker_callchain(id=PID)[-2].StartTime As ParentStartTime, 
+        if(condition=process_tracker_callchain(id=PID)[-2].EndTime<StartTime, then="", else=process_tracker_callchain(id=PID)[-2].EndTime) As ParentEndTime,
+        {SELECT Pid, Name, Ppid, Exe, CommandLine, Username, TokenIsElevated, StartTime,
```
scudette, line 71: You can use process_tracker_children() to get all the children of a process.

artifacts/definitions/Windows/Attack/UnexpectedImagePath.yaml
```diff
@@ -0,0 +1,73 @@
...
+        dllhost.exe,c:\windows\syswow64\dllhost.exe
+        wmiprvse.exe,c:\windows\system32\wbem\wmiprvse.exe
+        wmiprvse.exe,c:\windows\syswow64\wbem\wmiprvse.exe
+
+sources:
+  - precondition:
+      SELECT OS From info() where OS = 'windows'
+
+    query: |
+      LET expected_paths_tmp <= SELECT ProcName AS ProcessName FROM expected_paths
```
scudette, line 44: This query does not do anything.

artifacts/definitions/Windows/Attack/UnexpectedImagePath.yaml
```diff
@@ -0,0 +1,73 @@
...
+      LET expected_paths_lookup <= SELECT ProcessName AS ProcName, 
+        array(a1={SELECT ExpectedPath FROM expected_paths WHERE ProcessName=ProcName}) AS ExpectedPaths FROM expected_paths_tmp
+    
+      LET running_processes <= SELECT Pid AS PID, Name AS ProcessName, Ppid AS PPID, 
+        Exe AS ImagePath, CommandLine, Username, TokenIsElevated, StartTime, 
+        if(condition=EndTime<StartTime, then="", else=EndTime) AS EndTime
+        FROM process_tracker_pslist()
+        WHERE lowcase(string=Name) IN expected_paths.ProcName
```
scudette, line 52: What we want here is a lookup for proc name that will return a list of allowed paths quickly. You can use memoize() for extra speed although it probably wont matter in practice

artifacts/definitions/Windows/Attack/UnexpectedImagePath.yaml
```diff
@@ -0,0 +1,73 @@
...
+              WHERE ImagePath!="" AND lowcase(string=ProcessName)=ProcName AND NOT lowcase(string=ImagePath) IN ExpectedPaths
+          }
+      )
+
+      SELECT PID, ProcessName, ImagePath, CommandLine, Username, TokenIsElevated, StartTime, EndTime, 
+        PPID, process_tracker_callchain(id=PID)[-2].Data.Name As ParentProcessName, 
+        process_tracker_callchain(id=PID)[-2].Data.Exe As ParentImagePath, 
+        process_tracker_callchain(id=PID)[-2].Data.CommandLine As ParentCommandLine, 
+        process_tracker_callchain(id=PID)[-2].Data.Username As ParentUsername,
+        process_tracker_callchain(id=PID)[-2].Data.TokenIsElevated As ParentTokenIsElevated,
```
scudette, line 68: You can just lookup the parent process in the tracker - this will be more readable than call chain[-2] although it will also be very useful to actually report the call chain as well

artifacts/definitions/Windows/Attack/UnexpectedImagePath.yaml
```diff
@@ -0,0 +1,73 @@
...
+      )
+
+      SELECT PID, ProcessName, ImagePath, CommandLine, Username, TokenIsElevated, StartTime, EndTime, 
+        PPID, process_tracker_callchain(id=PID)[-2].Data.Name As ParentProcessName, 
+        process_tracker_callchain(id=PID)[-2].Data.Exe As ParentImagePath, 
+        process_tracker_callchain(id=PID)[-2].Data.CommandLine As ParentCommandLine, 
+        process_tracker_callchain(id=PID)[-2].Data.Username As ParentUsername,
+        process_tracker_callchain(id=PID)[-2].Data.TokenIsElevated As ParentTokenIsElevated, 
+        process_tracker_callchain(id=PID)[-2].StartTime As ParentStartTime, 
+        if(condition=process_tracker_callchain(id=PID)[-2].EndTime<StartTime, then="", else=process_tracker_callchain(id=PID)[-2].EndTime) As ParentEndTime,
```
scudette, line 70: Why do we check the endtime < start time all the time? it is not a good idea to change the types of the data randomly - better to return a NULL or indeed the endtime.

scudette (commented):
This is a very cool capability - we need to do it a bit more efficiently though.

Maybe something like this is a bit clearer?

```sql
      LET expected_paths_lookup <= memoize(key="ProcName", query={
         SELECT ProcName, enumerate(items=ExpectedPath) AS Path 
         FROM expected_paths
         GROUP BY ProcName
      })

      LET suspicious_processes = SELECT Pid AS PID, Name AS ProcessName, Ppid AS PPID, 
        Exe AS ImagePath, CommandLine, Username, TokenIsElevated, StartTime, 
        if(condition=EndTime<StartTime, then="", else=EndTime) AS EndTime,
        get(item=expected_paths_lookup, field=Name).Path AS ExpectedPaths,
        process_tracker_callchain(id=Pid) AS CallChain,
        process_tracker_get(id=Ppid) AS Parent
        FROM process_tracker_pslist()
        WHERE ImagePath != "" AND ExpectedPaths AND
          NOT lowcase(string=ImagePath) IN ExpectedPaths
        
      SELECT PID, ProcessName, ImagePath, CommandLine, Username, TokenIsElevated, StartTime, EndTime, 
        PPID, Parent.Data.Name As ParentProcessName, 
        Parent.Data.Exe As ParentImagePath, 
        Parent.Data.CommandLine As ParentCommandLine, 
        Parent.Data.Username As ParentUsername,
        Parent.Data.TokenIsElevated As ParentTokenIsElevated, 
        Parent.StartTime As ParentStartTime, 
        Parent.EndTime As ParentEndTime,
        CallChain.Data AS _CallChain,
        { SELECT Pid, Name, Ppid, Exe, 
                 CommandLine, Username, TokenIsElevated, StartTime, EndTime
          FROM 
          foreach(row=process_tracker_children(id=PID).Data)
        } AS SubProcesses
        FROM suspicious_processes
        
```

artifacts/definitions/Windows/Attack/UnexpectedImagePath.yaml
AmgdGocha: The reason which made me report only the parent and the child processes is these kinds of attacks usually include an injection in a process or maybe a malware that is persistent with some sort of a persistence mechanism, then the malware may spawn sub-processes to do its nefarious stuff, they usually don't include any sophistication in the call chain that may need to report more processes that this.

AmgdGocha:
Thanks, @scudette for this great review and explanation, I really appreciate your comments!
I've updated the artifact accordingly.

scudette (approved):

PR #2212, opened 2022-11-03
User admin

jeffmahoney:
This PR implements a user administration interface in the GUI. It requires some additional endpoints to list user roles, list organizations, get an individual user, create a user, update a user, and delete a user. The user interface is aware of organizations and does most of the right things -- but it follows the existing patterns where the system is _capable_ of allowing different users to have different roles in each org, but nothing actually does it yet.


jeffmahoney:
There's a lot here, but I'd like to focus on the API endpoints first since those will be permanent and I'd like to get this deployed locally soonish.

jeffmahoney:
user-admin-demo

jeffmahoney:
(No real changes with the re-push. There was just some oddities with generated pb files where they shouldn't be.)

api/users.go
```diff
@@ -53,35 +60,383 @@ func (self *ApiServer) SetPassword(
...
+	if in.WithRoles {
+		for _, user := range users {
+			user.Permissions, err = acls.GetPolicy(org_config_obj, user.Name)
+			if err != nil {
+				return nil, err
+			}
+		}
+	}
+
+	sort.Sort(userByName(users))
```
scudette, line 141: You can use sort.Slice() a bit easier here so you dont need another type.

api/users.go
```diff
@@ -53,35 +60,383 @@ func (self *ApiServer) SetPassword(
...
-	perm, err := acls.CheckAccess(org_config_obj, user_record.Name, permissions)
+	perm, err := acls.CheckAccess(org_config_obj, user_record.Name, acls.SERVER_ADMIN)
 	if !perm || err != nil {
 		return nil, status.Error(codes.PermissionDenied,
 			"User is not allowed to enumerate users.")
 	}
 
+	filterOrgs := true
+
+	if isMemberOfRootOrg(user_record.Orgs) {
```
scudette, line 93: There is an assumption here that a member of the root org will have admin access to all the other orgs which is not currently the case. It is maybe a reasonable assumption in practice though.

I think this function should replicate (or even refactor) the VQL equivalent here 
https://github.com/Velocidex/velociraptor/blob/master/vql/server/users/users.go#L77

Generally we should have once source of truth in the code and call it both from the VQL and API if we are extending the API.

api/users.go
```diff
@@ -53,35 +60,383 @@ func (self *ApiServer) SetPassword(
...
+	}
+
+	if password != "" {
+		users.SetPassword(user_record, password)
+	}
+
+	return nil
+}
+
+func grantUserRoles(config_obj *config_proto.Config, org_manager *services.OrgManager,
```
scudette, line 202: Should we just call the VQL function https://github.com/Velocidex/velociraptor/blob/master/vql/server/users/grant.go#L23

It makes it much easier to test and we then know the logic is the same as what is tested.

api/users.go
```diff
@@ -53,35 +60,383 @@ func (self *ApiServer) SetPassword(
...
+	if err == services.UserNotFoundError {
+		return nil, status.Errorf(codes.NotFound, "User %s does not exist.", in.Name)
+	}
+
+	org_admin, err := acls.CheckAccess(org_config_obj, active_user.Name, acls.ORG_ADMIN)
+	if err != nil {
+		return nil, err
+	}
+
+	if !org_admin && checkOrgMembership(user_record.Orgs, active_user.Orgs) == false {
```
scudette, line 396: You generally dont need to check org membership because the ACL exists in each org anyway so the check above ensures the user is an admin in the org.

api/users.go
```diff
@@ -53,35 +60,383 @@ func (self *ApiServer) SetPassword(
...
+				  in *api_proto.ChangeUserRequest) (*emptypb.Empty, error) {
+	return self.ChangeUser(ctx, in, false)
+}
+
+func (self *ApiServer) DeleteUser(
+	ctx context.Context,
+	in *api_proto.UserRequest) (*emptypb.Empty, error) {
+
+	users_manager := services.GetUserManager()
+	active_user, org_config_obj, err := users_manager.GetUserFromContext(ctx)
```
scudette, line 374: More standard to call this "principal" is the user that is initiating the activity

scudette:
Do you think it needs to go in the sidebar? The screen is only useful for admins and only sometimes (user management is not very common). We should try to keep the sidebar uncluttered as much as possible.

Maybe add a link from the welcome screen? Or maybe in the user preferences/change password screen?

Otherwise we can condition the sidebar entry on the user being an org admin.

scudette:
I noticed that protoc creates a bunch of churn in the repo and so pushed an update to the make_protobuf.sh - you should be able to resync and regenerate to clean up the proto modifications

jeffmahoney:
> I noticed that protoc creates a bunch of churn in the repo and so pushed an update to the make_protobuf.sh - you should be able to resync and regenerate to clean up the proto modifications

Thanks, I was wondering if there was a better way to do that.

api/users.go
jeffmahoney: Ack. Integrated.

api/users.go
jeffmahoney: I think this ends up being a fundamental issue with this entire PR. I wrote most of this against 0.6.4 before the organizations were introduced and had to update it against 0.6.6 (and now 0.6.7-rc1) to submit it. The user admin UI is not taking into account the current organization of the principal. It's displaying and operating in a global view, but constraining it by the list of orgs the principal is a member of and, when appropriate, an intersection of the orgs of the user being updated/deleted. I was basing that on  `VelociraptorUser`  having a list of orgs and a set of roles and mistakenly assuming that meant that the roles applied to every Org, when it probably means that the roles correspond to the current org.

I'll need to go back and rework this.

scudette: All you really need to do is to delegate to the VQL functions to do the actual work - the API should be a thin wrapper around the VQL which is supposed to do the right thing in all cases

api/users.go
jeffmahoney: I made an assumption early on that the root org was intended to be hierarchically higher than the other orgs, and now I see that assumption is invalid.

scudette: Kind of - for convenience an admin of the root org automatically gets the org admin permission as well
https://github.com/Velocidex/velociraptor/blob/master/acls/roles.go#L101

so we dont need a special role just for org admins but it is always possible to give users individual permissions and that permission is what actually matters

api/users.go
jeffmahoney: I suppose I don't have a clear grasp on what the roles mean in the scope of orgs and where the separation of the org takes effect.

I see that ORG_ADMIN in the current org allows:
- COLLECT_CLIENT permission in _any_ org via `v1/CreateHunt` and `hunt_add()`
- Listing users across orgs
- Creating new orgs

SERVER_ADMIN is typically evaluated against the current org, but often the operations the SERVER_ADMIN role can perform aren't limited to that org. For example, `user_grant()` checks for SERVER_ADMIN in the current org but then allows the caller to operate on roles in any org. `user_grant()` can also be used to modify the ACLs in each org separately so that having an ACL in one org doesn't mean it exists in other.

I suppose my code was trying to refine this a bit, at least for the API, so that SERVER_ADMIN only applies to the orgs to which the user belongs. ORG_ADMIN allows SERVER_ADMIN to be global, but the the direct implication of SERVER_ADMIN -> ORG_ADMIN means that distinction doesn't exist.

Do orgs only exist for the lesser privileged roles and SERVER_ADMIN is intended to be global? If that's the case, we should probably avoid inconsistencies with SERVER_ADMIN only being set on some orgs.

scudette: ORG_ADMIN always allows full org control. I guess it is an oversight that a SERVER_ADMIN can grant ORG_ADMIN to a user without having ORG_ADMIN themselves. 

This new permission model is pretty new so we do need to think through some of these implications. This is why it is important to have consistent calls into the same well tested set of functions instead of reimplement the logic in slightly different ways in different places.

If you dont mind I can help with the PR by refactoring the VQL functions to make them easier to call from the API functions. Otherwise this is really a great addition !  Thanks for kicking the tires on the permission system too, clearly we have some way to go on it :-).

The whole point of the multi org feature is to make it safe for an org admin to provision an org and give full admin rights (SERVER_ADMIN) to a user in that org without escalation to other orgs. One potential escalation path is arbitrary file write, execve and other dangerous plugins which #2214 is enforcing so we are making progress towards that goal - but we are not there yet!

jeffmahoney: I've already started in on refactoring the VQL stuff for user manipulation.  The general idea is to factor out fetching the ACLManager from VQL so that both the API and VQL callers can pass it in to a common implementation under services/users. I'm using a request struct for the refactor that gets replaced by the api_proto version later, so that there's no real difference on the VQL side and a straight line on the API side. In doing so I found a bit of a weakness in my original approach: you have to essentially specify the entire user with all the fields every time and when the gRPC http proxy ends up removing booleans if they're false or unset, they also always need to be specified. Not ideal. So now I'm using two maps of string->string and string->bool to indicate new values for fields.

scudette:
I have just submitted a bunch of user management functions that make it easier and safer to add/remove from orgs etc.

Looking at the UI in this PR I think it does not reflect the permission model in Velociraptor very well maybe we can rethink the layout ?

The model in Velociraptor is:
1. The server is divided into Orgs - they represent hard partitioning between servers. An admin for org A should never even see Org B exists. But at the same time should be free to add/remove users to their org (they do not need to be org admin) 
2. A user can belong to multiple orgs. Admins on multiple orgs can move users between the orgs.
3. Users have roles inside each org which can be independent (i.e. a reader in one org and admin in another org).

Common actions we need to perform with users:
1. Adding and removing a user from an org
2. Updating roles for a user inside an org.
3. Adding a new user
4. Adding a new org
5. Updating password for a user

I dont think this UI reflects this model:
* It seems that a user has the same roles in all orgs
* It is hard to see which users are in each org because the UI shows the orgs for each user (I think this is less useful view). I think most of the time this is the view people want to know most - who has rights within their org.

Maybe a tree view is better? Have the top level being the org, the under each org we see the users, then one can click on each user to see their roles within the org.

Alternatively maybe a cascading table with two levels? Left side table is orgs when user selects the org, next table populates with users in the org, then when a user is selected we show the roles etc. 

Maybe then have a toolbar or something to add the user to other orgs (with a required role) so then the same user shows up under different orgs?

jeffmahoney:
I think most of what I have in the UI now can be repurposed to work with orgs properly. I think I would represent the users by separating the user list on the right into three areas: users belonging to the active org, next users belonging to orgs other than the current one but in which the user is an admin and can still see full user records, and last users belonging to other orgs entirely.

While looking through the code, I think there is still some ambiguity as to how the org admin role is supposed to work. In some uses it's only checked against the root org but in others it's only checked against the current org. Is the org admin role intended to be global or is it intended to have the ability to add and remove users only in the orgs where the user holds that role?

scudette:
The org admin role is basically a bootstrap mechanism to allow orgs to be created in the first place. Once orgs have their own admin the intention is that they should be able to manage all their own users. I have not thought too much about the implications of making a user have org admin in the org rather than the root org - it might only make sense to have org admin in the root org anyway. Currently the permission is provided by the "administrator" role only in the root org.

We probably should also stop an org admin from giving the permission on the org to another user too.

Note that the list of users is actually filtered now to only those users that the principal has visibility of (i.e. if they have a reader permission in the org they can only see their name, admin perm in any of their org will see their full details).

Normal use case is that an admin has one or two orgs and some users  already in these orgs. They might need to move users between the orgs or create new users but have no ability to bring other users into the org at all (they can not see them even). So the third category in your GUI will always be empty.

jeffmahoney:
Ok, sure, the UI won't have it because ListUsers won't, but that's the only reason. Both the API and the VQL `user_create()` plugin will allow a user to add any user to their org as long as they have the server admin role in the orgs they want to add users. I suppose that's not too much of an inconsistency since the the VQL/API would work the same and the caller has to know the username.

jeffmahoney:
As I'm figuring out the API for updating users correctly using orgs as first-class citizens in the system, I started wondering if the ORG_ADMIN role needs to be refined some more or dropped.

- We talked about how it probably doesn't make sense to have ORG_ADMIN in a non-root org.
- ORG_ADMIN currently functions as a global SERVER_ADMIN role. It technically has visibility into every org. It can create users and add/remove them from any org. It can define permissions for any user in any org, including itself, so it effectively has SERVER_ADMIN in every org. Even if it didn't, it could create a user with privilege and use that user to give itself privileges.
- You'd mentioned that you viewed ORG_ADMIN as a way to bootstrap orgs, but I suppose it's also a way to recover orgs if an org ends up somehow with no SERVER_ADMIN users left.

Is there a use case for a user to have SERVER_ADMIN in the root org but not ORG_ADMIN?
Is there a use case for a user to have ORG_ADMIN in the root org but not SERVER_ADMIN?

If the answers to both of these are "no" we should drop ORG_ADMIN entirely. If there is a possible "yes" then we need to determine what the distinction is and not automatically imply ORG_ADMIN from SERVER_ADMIN in the root org.

To a certain extent, I think orgs should either be "off" and there's only the root org or "on" and the root org only functions to provide a context outside of the orgs that can be used to manage them. In the latter case,  the privileges beyond ORG_ADMIN/SERVER_ADMIN aren't relevant anymore.

jeffmahoney:
Here's what I've got now in terms of UI. I need to split up the changes and stuff them back into easily digestible commits, but it works. The switches select orgs and once selected, the appear in the accordion. Each org title is a link that displays the roles for that org as well as the current effective permissions. The top of the user select pane will always contain the active current org. If the user has access to other orgs, they'll be displayed below. Likewise with org admins. 

image

scudette:
> As I'm figuring out the API for updating users correctly using orgs as first-class citizens in the system, I started wondering if the ORG_ADMIN role needs to be refined some more or dropped.
> 
> * We talked about how it probably doesn't make sense to have ORG_ADMIN in a non-root org.
> * ORG_ADMIN currently functions as a global SERVER_ADMIN role. It technically has visibility into every org. It can create users and add/remove them from any org. It can define permissions for any user in any org, including itself, so it effectively has SERVER_ADMIN in every org. Even if it didn't, it could create a user with privilege and use that user to give itself privileges.

Corrent - the idea is that ORG_ADMIN is like a super user for the entire deployment and is used to bootstrap individual orgs.

> * You'd mentioned that you viewed ORG_ADMIN as a way to bootstrap orgs, but I suppose it's also a way to recover orgs if an org ends up somehow with no SERVER_ADMIN users left.

This is very easy to do since there is no logical check in user_delete() to make sure there are any admins left over - currently this allows deleting the only admin user for the org leaving the org orphaned :-). It is also possible to remove ORG_ADMIN in the same way which requires re-adding the user from the command line. I was debating if we should build some foot/gun protections here but at least the ORG_ADMIN can fix things if they happen.

> Is there a use case for a user to have SERVER_ADMIN in the root org but not ORG_ADMIN? Is there a use case for a user to have ORG_ADMIN in the root org but not SERVER_ADMIN?

An important clarification here is that ORG_ADMIN is a permission, like SERVER_ADMIN is a permission. A role can contain multiple permissions - so the "administrator" role contains both permissions. For clarity it is always better to have clearly defined permissions - even if they typically go together or in practice one implies the other.  The ORG_ADMIN and SERVER_ADMIN permissions are very different actions I think so it is still worth separating these out. However, I do acknowledge that they typically go together (at least for the root org) and so the "administrator" role can provide both automatically.

> If the answers to both of these are "no" we should drop ORG_ADMIN entirely. If there is a possible "yes" then we need to determine what the distinction is and not automatically imply ORG_ADMIN from SERVER_ADMIN in the root org.

We don't imply these permissions are the same - we just provide a convenient role which grants both these permissions. It is possible to not have that permission at all. So in theory it is possible to have an admin for the root org which does not have ORG_ADMIN permission at all.

> To a certain extent, I think orgs should either be "off" and there's only the root org or "on" and the root org only functions to provide a context outside of the orgs that can be used to manage them. In the latter case, the privileges beyond ORG_ADMIN/SERVER_ADMIN aren't relevant anymore.

I'm not sure I understand this statement :-). Are you suggesting that "administrator" role should not provide ORG_ADMIN by default? so going from a single org deployment to a multi org one requires first granting an additional role?

scudette:
> Here's what I've got now in terms of UI. I need to split up the changes and stuff them back into easily digestible commits, but it works. The switches select orgs and once selected, the appear in the accordion. Each org title is a link that displays the roles for that org as well as the current effective permissions. The top of the user select pane will always contain the active current org. If the user has access to other orgs, they'll be displayed below. Likewise with org admins.
> 
> image

This looks cool but I wonder how it will scale to many users and many orgs. I wonder if a tree style hierarchy is more understandable here? I mean in this screenshot the third column from the left is already overflowing and there are only 2 users in 1 org.

Perhaps we need two types of UI - to view the roles/permissions we use a simple compressed widget (so in this case it just shows "Read Only User" and does not show all the other "off" roles) and then a GUI that allows to update roles/permissions with a more detailed view of all permissions (maybe in a modal?)

From a task based approach the user might want to do the following workflows:
1. See all the users assigned to an org and their roles
2. See all the orgs a user has access to.
3. Modify a user's information (password etc).

The first requires drilling down: first all the orgs, then selecting the org, all the users in the org, then selecting the user all the roles for that user.
The second workflow requires drilling down the user, then viewing all the orgs the user belongs to and then all the roles they have in these orgs.
The third workflow is really not related to the first 2 at all so maybe we should just add it in a modal toolbar when the user selects a user? This will remove the first column from the real estate - since it does not really add value for the first 2 workflows it should not consume extra space. 

It is almost as though we need two separate GUIs:
1. First column is user search, then click on user, second column is org search, click on org view roles
2. First column is org search, click on org, second column is user search, then roles.

Then have the user switch between them (e.g. with tabs) - view by user, view by org

jeffmahoney:
> > * You'd mentioned that you viewed ORG_ADMIN as a way to bootstrap orgs, but I suppose it's also a way to recover orgs if an org ends up somehow with no SERVER_ADMIN users left.
> 
> This is very easy to do since there is no logical check in user_delete() to make sure there are any admins left over - currently this allows deleting the only admin user for the org leaving the org orphaned :-). It is also possible to remove ORG_ADMIN in the same way which requires re-adding the user from the command line. I was debating if we should build some foot/gun protections here but at least the ORG_ADMIN can fix things if they happen.
> 
> > Is there a use case for a user to have SERVER_ADMIN in the root org but not ORG_ADMIN? Is there a use case for a user to have ORG_ADMIN in the root org but not SERVER_ADMIN?
> 
> An important clarification here is that ORG_ADMIN is a permission, like SERVER_ADMIN is a permission. A role can contain multiple permissions - so the "administrator" role contains both permissions. For clarity it is always better to have clearly defined permissions - even if they typically go together or in practice one implies the other. The ORG_ADMIN and SERVER_ADMIN permissions are very different actions I think so it is still worth separating these out. However, I do acknowledge that they typically go together (at least for the root org) and so the "administrator" role can provide both automatically.
> 
> > If the answers to both of these are "no" we should drop ORG_ADMIN entirely. If there is a possible "yes" then we need to determine what the distinction is and not automatically imply ORG_ADMIN from SERVER_ADMIN in the root org.
> 
> We don't imply these permissions are the same - we just provide a convenient role which grants both these permissions. It is possible to not have that permission at all. So in theory it is possible to have an admin for the root org which does not have ORG_ADMIN permission at all.

Right, and that's the distinction I'm trying to draw. In theory it's possible to separate them but we're not. If it's defined as not having to be implied, and that it's just an implementation detail, I can deal with that.

> > To a certain extent, I think orgs should either be "off" and there's only the root org or "on" and the root org only functions to provide a context outside of the orgs that can be used to manage them. In the latter case, the privileges beyond ORG_ADMIN/SERVER_ADMIN aren't relevant anymore.
> 
> I'm not sure I understand this statement :-). Are you suggesting that "administrator" role should not provide ORG_ADMIN by default? so going from a single org deployment to a multi org one requires first granting an additional role?

I'm saying that when there is a single org, the root org is not special. It just is the only org.  When there are multiple orgs, and we keep ORG_ADMIN in the root org, it makes the root org special in terms of how orgs and accounts are managed. Yet it still remains a regular org that interacts with clients/etc in the same way as other orgs. I'm saying we should codify that such that the root org in a multi-org environment is special and the only privileges that can be granted are SERVER_ADMIN and ORG_ADMIN. Perhaps when the second org is created, the first should be converted to a new one as well.


 but not in how it interacts with clients/etc.  It stops being a "peer" org and ends up being kind of a "global" org  since a user with ORG_ADMIN _must_ be in the root. Given that distinction, I think it would make sense to define the root org as special in a multi-org environment and to either define the system as being multi-org from the beginning or to migrate the root org to a "new" org when the second org is created.

scudette:
I have reworked the GUI to work more along the most common workflow:

1. In the users tab start by selecting a user, see what orgs they belong in
2. Click on any org will show the roles/permissions
3. Can add/remove roles from that org for that user
4. Can add extra permissions by clicking on the effective permissions section
5. Can remove extra permissions
6. When all the roles/permissions are removed a dialog asks if we really want to delete the user (this is effectively deleting the user from that org because they have no permissions in that org).
7. Can add a new user to that org

Velociraptor_Response_and_Monitoring

We need to add a username/password reset section still.

PR #2363, opened 2023-01-03
Create Server.Alerts.ProcessCreation.yaml

DfirJos:


DfirJos (commented):
I've renamed the client_event artifact.

PR #2497, opened 2023-03-05
Refactoring side panel navigation as "main menu" navigation, tweaked the hamburger button

azemetre:
Hello!

I did a few tweaks to improve the accessibility of the sidebar.

I created a new landmark "main menu" navigation (see screenshots to see the difference).

In the `<VeloNavigator />` I opted to remove the ternary that renders the "disabled" links as spans or elements and opted to just render links then hide it from assistive tech using `aria-hidden`. There was a slight challenge in ensuring the links don't get keyboard focus which is why I have the `tabIndex` as well.

My only question with this is what are the possible values of `disabled` in the `<VeloNavigator />` component? I'm seeing null and disabled, but couldn't find all possible values. Hopefully the null check is sufficient enough?

There is another approach in utilizing the `aria-disabled` but it requires tweaking the `hrefs` appropriately to prevent a user from accidentally going to a page that is disabled, once this happens all the main menu navigation links become enabled. This would seem more useful for disabling form controls like `<input />`, `<button />`, or `<textarea />`.

Added a few more phrases to the i18n dictionaries, then used where appropriate.

If you notice in the screenshots it says "list 8 items" but reads the enabled links as "3 out of 12," this is fine in my experience. As users get familiar with using the application consistency is useful. When a user learns how to enable the "Host Information," "Virtual Filesystem," "Collected Artifacts," and "Client Events" links, navigation to the main menu and jumping to the relevant link becomes easier.

The hamburger span is now a hamburger button, if you take a peak at the screenshots you'll see another aria attribute `expanded` lets users know what is happening as they are activating the button.

Other minor tweaks was removing the duplicate `alt` text in the `sidebar_links`, the `T(x.text)` is already sufficient otherwise it repeats the information twice in a screenreader for example.

<details><summary>voiceover screenshots before/after</summary>

Previous iteration:

previous-main-menu-naviation

previous-main-menu-link

New Iteration:

button-toggle-collapsed

button-toggle-expanded

main-menu-navigation-landmark

main-menu-list-active-link-number

main-menu-list-link-out-of-total

main-menu-list-current-link

Hardcoded a disabled link as enabled for reference:

main-menu-list-virtual-filesystem
</details>

### DIFF
--- a/gui/velociraptor/src/components/i8n/de.jsx
+++ b/gui/velociraptor/src/components/i8n/de.jsx
@@ -96,6 +96,9 @@ const Deutsch = {
     "Create a new artifact": "Erstellen eines neuen Artefakt",
     "Save": "Speichern",
     "Search": "Suchen",
+    "Toggle Main Menu": "Hauptmenü umschalten",
+    "Main Menu": "Hauptmenü",
+    "Welcome": "Willkommen",
 
     // Keyboard navigation.
     "Global hotkeys": "Globale Hotkeys",
--- a/gui/velociraptor/src/components/i8n/en.jsx
+++ b/gui/velociraptor/src/components/i8n/en.jsx
@@ -46,6 +46,9 @@ const English = {
     "Shell": "Shell",
     "Close": "Close",
     "Search": "Search",
+    "Main Menu": "Main Menu",
+    "Toggle Main Menu": "Toggle Main Menu",
+    "Welcome": "Welcome",
     "Connected": "Connected",
     "time_ago": (value, unit) => {
         return value + ' ' + unit + ' ago';
--- a/gui/velociraptor/src/components/i8n/es.jsx
+++ b/gui/velociraptor/src/components/i8n/es.jsx
@@ -97,6 +97,9 @@ const Spanish = {
     "Create a new artifact": "Crear un nuevo artefacto",
     "Save": "Guardar",
     "Search": "Buscar",
+    "Toggle Main Menu": "Alternar menú principal",
+    "Main Menu": "Menú principal",
+    "Welcome": "Bienvenido",
 
     // Keyboard navigation.
     "Global hotkeys": "Teclas de acceso directo globales",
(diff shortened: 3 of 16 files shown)


scudette:
Thanks for this PR ! This is very cool. There seem to be a slight change in layout though as illustrated in the below comparison

Win_Server_2019_2_and_Microsoft_Remote_Desktop

Previously the hamburger took the full line height but now the icon is kind of on the bottom and takes about half the height.  I think it looks better as a button that takes the full height like the button next to it. Can we adjust the height somehow?

azemetre:
> Thanks for this PR ! This is very cool. There seem to be a slight change in layout though as illustrated in the below comparison
> 
> Win_Server_2019_2_and_Microsoft_Remote_Desktop
> 
> Previously the hamburger took the full line height but now the icon is kind of on the bottom and takes about half the height. I think it looks better as a button that takes the full height like the button next to it. Can we adjust the height somehow?

Sure thing! Just to confirm you're referring to the spacing between the black bars correct?

Would it also be possible to know the browser and OS you are using? I'm on chrome with mac os but I use accessibility settings myself like increasing the contrast; on chrome the spacing between the bars is identical. I just wonder if the unicode character being used (TRIGRAM FOR HEAVEN) is getting rendered as another similar unicode character (IDENTICAL TO), I'll have to investigate more to match the image. I'll also play around with the `<FontAwesomeIcon />` as I see they have some menu bar icons as well.

scudette:
I apologize - this does not appear related to this PR - I see a difference between the rendering of the icon on MacOS chrome and Windows Chrome (and Edge). The left example I posted was macos chrome and the right was edge (just verified with chrome same). Further this does not seem related to this PR, it shows the same problem in master branch as well

I wonder if the issue is the differences in rendering the unicode font between the different OSs. Maybe it is best to use the Font Awesome icon instead (called "bars") https://fontawesome.com/icons/bars?s=solid&f=classic This way we will have a consistent rendering between operating systems/browser versions.

scudette:
Lets merge this one and fix the icon to use Font Awesome in another PR, since this layout issue is not related to this PR

PR #2698, opened 2023-05-22
S3_Upload: Adding KMS and Prefix arguments

superbobdthm:
Changes
-------
- Adds argument `kmsEncryptionKey` to allow for specifying the KMS key to be used with the upload
- Adds argument `S3UploadRoot` to allow for specifying a prefix to apply to any file names uploaded.


superbobdthm:
rebasing

PR #2766, opened 2023-06-21
Audit user creation and user role modifications.

### LINKED ISSUE #2761: Improve API Auditing
Some API-backed actions such as user creation or ACL changes are not audited within the `Server.Audit.Logs` artifact while some VQL-backed counterparts (e.g.: user_create) are. All administrative actions should be audited regardless of their origin (VQL, API, CLI?, ...).

- [X] VQL-backed user creation is logged at https://github.com/Velocidex/velociraptor/blob/d9a830480253cfa796184895284ce7b25f71a0ba/vql/server/users/create.go#L70
- [ ] API-backed user creation should be logged at https://github.com/Velocidex/velociraptor/blob/d9a830480253cfa796184895284ce7b25f71a0ba/api/users.go#L127
- [ ] User addition to organizations and ACL changes should be logged at https://github.com/Velocidex/velociraptor/blob/d9a830480253cfa796184895284ce7b25f71a0ba/api/users.go#L247

0xThiebaut:
This PR introduces the `user_create` API audit operation using the same format as the audit entry generated through the VQL `user_create` function. A new `user_grant` audit operation is furthermore introduced to audit role changes performed both through the API and VQL `user_grant` function. The `user_grant` audit entry has the same structure as the `user_create` audit entry.

Closes #2761 

image

---

This PR furthermore improves build dependency documentation and makes the GUI Makefile Windows-compatible.

When running `make` on Windows, the `'./node_modules/go.mod'` path is not found when using quotes.

```
PS velociraptor\gui\velociraptor> make build
echo "module fake_nodejs_module // Exclude this directory from Go tools\n\ngo 1.16" > './node_modules/go.mod'
The system cannot find the path specified.
make: *** [Makefile:6: build] Error 1
```

Furthermore, the `mkdir` command's `-p` flag is neither Windows compatible, nor required as `npm run build` will generate the `build` directory. The path seprator also differs between Linux and Windows, hence requiring a preliminary `cd`.

```
PS velociraptor\gui\velociraptor> make build
[...]
mkdir -p build/static/
The syntax of the command is incorrect.
make: *** [Makefile:8: build] Error 1
```

By removing the quotes (as done with other commands within the Makefile) and incompatible `mkdir` command flag, `make` succeeds as expected.


api/users.go
```diff
@@ -125,6 +126,14 @@ func (self *ApiServer) CreateUser(ctx context.Context,
 	}
 
 	err = users.AddUserToOrg(ctx, mode, principal, in.Name, in.Orgs, acl)
+
+	services.LogAudit(ctx,
```
scudette, line 130: should we log only if err == nil?

api/users.go
```diff
@@ -246,5 +255,13 @@ func (self *ApiServer) SetUserRoles(
 	// users.AddUserToOrg
 	err = users.AddUserToOrg(ctx, users.UseExistingUser,
 		principal, in.Name, []string{in.Org}, acl)
+
+	services.LogAudit(ctx,
+		org_config_obj, principal, "user_grant",
+		ordereddict.NewDict().
+			Set("Username", in.Name).
+			Set("Roles", in.Roles).
```
scudette, line 263: we probably should log all the permissions as well because the call could include additional permissions which are not roles.

vql/server/users/grant.go
```diff
@@ -72,6 +73,13 @@ func (self GrantFunction) Call(
 		return vfilter.Null{}
 	}
 
+	services.LogAudit(ctx,
```
scudette, line 76: Should we only log if successful?

vql/server/users/grant.go
```diff
@@ -72,6 +73,13 @@ func (self GrantFunction) Call(
...
+		org_config_obj, principal, "user_grant",
+		ordereddict.NewDict().
+			Set("Username", arg.Username).
+			Set("Roles", arg.Roles).
```
scudette, line 80: We should probably also log the policy.

This function actually can be used by either providing roles or a policy (which might include roles plus extra permissions)

scudette (changes requested):

vql/server/users/grant.go
0xThiebaut: Updated the PR to only log successful events.

0xThiebaut:
Resolved the above comments, standardized the keys to match the VQL parameter format and further introduced organization-related audit events. This PR is hence to be considered a breaking change on audit logs.

image

PR #2768, opened 2023-06-21
Expose user policy through `gui_users` for use with `user_grant`

0xThiebaut:
Exposes the user policy through `gui_users`, enabling the output to be used with `user_grant` without relying on the effective ACLs.


scudette (commented):
The plugin does expose the policy with the _policy column. It is preceded with an _ to hide it from the GUI by default but an artifact can easily access that

0xThiebaut:
Indeed, just made the PR as it wasn't intuitive to me the field existed when doing a `SELECT * FROM gui_users()` in a notebook. Feel free to close the PR if you'd rather keep the current behavior :)

PR #3039, opened 2023-10-18
linux trash dir artifact to find recycled files

shortstack:
find all ~/.local/share/Trash dir contents on nix systems and display output

### DIFF
--- a/artifacts/definitions/Linux/Sys/Trash.yaml
+++ b/artifacts/definitions/Linux/Sys/Trash.yaml
@@ -1,0 +1,24 @@
+name: Linux.Sys.Trash
+description: Find files in trash bin for all users
+author: Whitney Champion (@shortxstack)
+parameters:
+  - name: trashFiles
+    default: '.local/share/Trash/**/*'
+    description: Glob of trash bin files relative to a user's home directory.
+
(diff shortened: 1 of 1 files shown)


artifacts/definitions/Linux/Sys/Trash.yaml
```diff
@@ -0,0 +1,24 @@
...
+  - precondition: |
+      SELECT OS From info() where OS = 'linux'
+
+    query: |
+      LET trash = SELECT * from foreach(
+          row={
+             SELECT Uid, User, Homedir from Artifact.Linux.Sys.Users()
+          },
+          query={
+             SELECT FullPath, Mtime, Ctime, User, Uid
```
mgreen27, line 19: Could you switch this to OSPath? Is it worth putting a Size field too?

shortstack: fixed/added :)

scudette:
I think this is covered by this rule

https://triage.velocidex.com/docs/linux.triage.uac/rules/#Trash_info

although that rule only grabs the trash info - do you want to also include all the files themselves in the trash dir?

PR #3050, opened 2023-10-25
frontend: Detect, log duplicate minion names

hillu:
If two or more minions reported metrics using the same name, they would just overwrite each other, leading to only the last set of metrics in each 10s interval to be reported via the Prometheus interface.


hillu:
You have removed this patch as part of c1903af4241f4000ba9c69114aced43452a270fd; I am guessing by mistake.

Issue #2290, opened 2022-12-01
Exporting a collection should also export the notebook


scudette:
Importing the collection again should also import the notebook as well.

PR #3086, opened 2023-11-09
Export notebooks with exported collections #2290

bmcder02:
PR for issue #2290 , adding notebooks to the list of items exported when creating a download file.


bmcder02:
Leaving in draft for the moment. Flows are exporting the notebooks fine. Hunts are exporting the notebooks as empty files, currently troubleshooting.

PR #3220, opened 2024-01-19
Update LastUserLogin.yaml

CyberSecUniCorn:
* Combine Login/Logout into one line in order to determine the session length more easy.


mgreen27:
looks good - thank you!

PR #3260, opened 2024-01-31
Bugfix: specify monospace font explicitly. #3259

fukusuket:
## What Changed
- Fixed: #3259
  - I have applied the following workaround.
    - https://github.com/ajaxorg/ace/issues/2548#issuecomment-699484043

### Enviroment
- OS: macOS Sonoma version 14.2.1
- Velociraptor: 0.7.1

### Test(Google Chrome)
I run the steps to reproduce #3259 and confirmed that the issue was resolved as follows.
スクリーンショット 2024-02-01 5 59 58

### Test(Firefox)
スクリーンショット 2024-02-01 6 05 34

### Test(Safari)
スクリーンショット 2024-02-01 6 10 09

Thank you for your time.
Best,

### DIFF
--- a/gui/velociraptor/src/themes/ncurses.css
+++ b/gui/velociraptor/src/themes/ncurses.css
@@ -141,6 +141,10 @@ body.ncurses {
   font-size: var(--font-size-base) !important;
 }
 
+.ncurses .ace_editor, .ace_editor * {
+  font-family: "Monaco", "Menlo", "Ubuntu Mono", "Droid Sans Mono", "Consolas", monospace !important;
+}
+
 /* this is for the context help popup */
 .ncurses .ace_editor {
   font-family: var(--font-family-monospace) !important;


fukusuket:
### Enviroment
- OS: Windows11
- Velociraptor: 0.7.1

### Test(Google Chrome)
chrome

### Test(Microsoft Edge)
ms

gui/velociraptor/src/themes/ncurses.css
```diff
@@ -141,6 +141,10 @@ body.ncurses {
   font-size: var(--font-size-base) !important;
 }
 
+.ncurses .ace_editor, .ace_editor * {
```
scudette, line 144: Probably we dont want to add to this file since this is only for the ncurses theme.

The font family is already set in the theme file for example
https://github.com/Velocidex/velociraptor/blob/00cbd2ff1c210a5a7c4c88ea864feba08472e1cf/gui/velociraptor/src/themes/veloci-light.css#L45

We actually include fonts already so the browser should not need to fall back to system fonts at all. 

Can you please check if you see the same problem in all themes and then use the dev console to check what font is actually loaded at the position of the caret?

PR #3438, opened 2024-04-21
Theme fixes and multichoice selector styling

predictiple:
* Added multichoice selector to Demo.Plugins.GUI and associated styling.
* Changed default ACE font size so that it's not disproportionately large when used in form controls:
Screenshot from 2024-04-21 11-41-02
* Styled ACE "suggestions" / placeholder text
* Miscellaneous CSS fixes and improvements in all the themes, particularly the new json viewer component and making the Midnight theme "more Metal" :-)


scudette:
This change moves the tooltip to the bottom of the suggestion box rather than next to the highlighter suggestion - is this what we want? It is hard to see the suggestion is related to the highlight because they are so far apart (your screenshot shows it the way it was before)

image

predictiple:
That's strange. On mine the tooltip is still off to the left side where it belongs.
Screenshot from 2024-04-21 16-26-56

I can't see anything that would have changed it either.

scudette:
Sorry disregard - this was my mistake - it puts the tooltip at the bottom when the browser is really narrow

PR #3441, opened 2024-04-22
Updated artifacts filter

predictiple:
- Replaced "Custom" regex with "Sigma" since custom rules are more accurately filtered by the "Custom Only" filter (i.e. builtin:no).
- Replaced OSX filter label with macOS since this is Apple's currently preferred terminology.


predictiple:
It would be cool if we could filter for "Basic" artifacts. I see there is some existing code for hidden artifacts but obviously you can't show something that's hidden.

PR #3455, opened 2024-04-26
Update to CuratedSigma artifact and some GUI improvements

predictiple:
- CuratedSigma artifact can now import multiple sources
- Improved the react-select component's appearance in all themes. Change also to it's behaviour so that it closes after each selection because if it stays open then it overlays the parameter control below it, which is particularly confusing when only 1 item remains to be selected in the list.

### DIFF
--- a/artifacts/definitions/Server/Import/CuratedSigma.yaml
+++ b/artifacts/definitions/Server/Import/CuratedSigma.yaml
@@ -27,14 +27,13 @@ parameters:
 
 sources:
   - query: |
-      LET PackageNames <= parse_json_array(data=PackageNames)
       LET URLlookup = dict(
         `Velociraptor ChopChopGo Ruleset (Linux)`="https://sigma.velocidex.com/Velociraptor-ChopChopGo-Rules.zip",
         `Velociraptor Hayabusa Ruleset`="https://sigma.velocidex.com/Velociraptor-Hayabusa-Rules.zip",
         `Velociraptor Hayabusa Live Detection`="https://sigma.velocidex.com/Velociraptor-Hayabusa-Monitoring.zip")
 
       SELECT * FROM foreach(row=PackageNames,
-                            query={SELECT * FROM 
+                            query={SELECT * FROM
                                 Artifact.Server.Import.ArtifactExchange(
                                 Prefix=Prefix,
                                 ExchangeURL=get(item= URLlookup, member= _value))})
--- a/artifacts/definitions/Server/Import/CuratedSigma.yaml
+++ b/artifacts/definitions/Server/Import/CuratedSigma.yaml
@@ -14,7 +14,7 @@ required_permissions:
 parameters:
   - name: PackageNames
     type: multichoice
-    default: Velociraptor Hayabusa Ruleset
+    default: '["Velociraptor Hayabusa Ruleset"]'
     choices:
       - Velociraptor Hayabusa Ruleset
       - Velociraptor Hayabusa Live Detection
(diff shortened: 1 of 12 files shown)


artifacts/definitions/Server/Import/CuratedSigma.yaml
```diff
@@ -22,15 +22,19 @@ parameters:
       - Velociraptor Curated Windows Ruleset
 
   - name: Prefix
-    description: Add artifacts with this prefix
+    description: Add this prefix to imported artifacts
+    validating_regex: '^[a-zA-Z0-9_.]$'
 
 sources:
   - query: |
+      LET PackageNames <= parse_json_array(data=PackageNames)
```
scudette, line 30: Is that needed? the compiler should add that in. If not we should add it to compiler.go

gui/velociraptor/src/components/forms/form.jsx
```diff
@@ -449,7 +449,7 @@ export default class VeloForm extends React.Component {
                       placeholder={T("Choose one or more items")}
                       className="velo"
                       classNamePrefix="velo"
-                      closeMenuOnSelect={false}
+                      closeMenuOnSelect={true}
```
scudette, line 452: I think this will be annoying- when we leave the menu open we can just select multiple items quickly otherwise we will need to click again to open it each time .

For example the category selector in the registry hunter

artifacts/definitions/Server/Import/CuratedSigma.yaml
predictiple: yes it's currently needed. The multichoice selctor presents a string.

scudette: Ok lets fix it :-)

gui/velociraptor/src/components/forms/form.jsx
predictiple: Yes I agree. I just tried it with RegistryHunter and there it's definitely better that it stays open. I hadn't considered such long selection lists. I'll revert it.

artifacts/definitions/Server/Import/CuratedSigma.yaml
```diff
@@ -12,8 +12,8 @@ required_permissions:
 - SERVER_ADMIN
 
 parameters:
-  - name: PackageName
-    type: choices
+  - name: PackageNames
+    type: multichoice
     default: Velociraptor Hayabusa Ruleset
```
predictiple, line 17: @scudette Do you think there should be a default value? The multichoice selector doesn't show the default value so it might be better to have nothing as default.

scudette: It should show the default value - it needs to be encoded as json array though because that is what the selector uses

predictiple: Ok thanks. I've updated the paramter and tested using only the default value and it seems fine.

artifacts/definitions/Server/Import/CuratedSigma.yaml
predictiple: Very strange. I can't see why it's not matching in compiler.go
Screenshot from 2024-04-26 14-51-56

scudette: I dont think this case is handled at all - we need to add it there

predictiple: Should this not handle it?: https://github.com/Velocidex/velociraptor/blob/c2fd3c4898039c491e894fd236abcdfbd71d5c85/services/launcher/compiler.go#L145

scudette: Almost - that case statement applies for "json_array", "regex_array" but we need to also apply it to multichoice which uses json array under the hood (but the GUI does something else with it)

predictiple: Thanks, I understand now.

PR #3509, opened 2024-05-22
Fix file path in Windows.Sigma.EventLogs artifact

predictiple:


predictiple:
The "upload" artifact paramter seems to have problems with wss. Works fine when it's https though.
Screenshot from 2024-05-22 23-31-09

PR #3626, opened 2024-07-17
Add artifact for installing and configuring deb packages

misje:
In Debian-based OSes software is preferably installed through the package manager. This artifact provides a comfortable way to install packages, as well as configuring them using debconf. The package can either be a name or a deb file, which can be optionally uploaded.

The purpose is to provide a reusable component for installing and configuring tools and other packages needed by other artifacts in a single command. `apt-get update` is optionally run, and dependencies, if any, is automatically installed. debconf may be pre-seeded with answers if the package requires configuration.

Other artifacts can offload all installation and configuration logic to this artifact by calling it like this: `SELECT * FROM Artifact.Linux.Utils.InstallDeb(DebName='sl')` or `SELECT * FROM Artifact.Linux.Utils.InstallDeb(DebName='/tmp/mypackage_0.1.0-1_amd64.deb')`. A row with *Step* "Installing package" and *ReturnCode* 0 will indicate that the installation was a success.

### DIFF
--- a/artifacts/definitions/Linux/Utils/InstallDeb.yaml
+++ b/artifacts/definitions/Linux/Utils/InstallDeb.yaml
@@ -1,0 +1,175 @@
+name: Linux.Utils.InstallDeb
+author: Andreas Misje – @misje
+description: |
+   Install a deb package and configure it with debconf answers. The package
+   may either be specified by name or be an uploaded file. If the package
+   already exists, it may be optionally reconfigured with debconf answers.
+
+type: CLIENT
+
+reference:
+   - https://manpages.debian.org/bookworm/debconf-doc/debconf-devel.7.en.html#Type
+
+parameters:
+   - name: DebName
+     description: |
+        Package to install (by name). Ignored if DebFile is set. An absolute path
+        to a deb file that already exists on the system is also accepted.
+
+   - name: DebFile
+     description: |
+        Package to install (by file). Remember to click "Upload"! When set,
+        DebName is ignored. Use DebName with an absolute file path if the
+        file already exists on the system and does not need to be uploaded.
+     type: upload
+
+   - name: UpdateSources
+     description: |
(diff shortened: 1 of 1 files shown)


scudette:
This very powerful artifact should probably require admin permissions to run.

misje:
Any suggestions on what permission that should be? I looked through the permissions list and and use of "required_permissions" in all other artifacts and couldn't find anything fitting.

How would this new permission potentially affect artifacts like those that already exist, relying on tools not installed in Linux by default, if they had been written to use something like this artifact to optionally install said tool automatically? For instance, Linux.Events.TrackProcesses installs auditd if selected. Otherwise you'd have to run another hunt to install it separately, which is not very user-friendly.

scudette:
Well considering that the artifact allows users to push and run arbitrary code probably EXECVE should be required.

misje:
That I agree, and I was considering just that.

misje:
Just to confirm: Permission requirements are not inherited when an artifact calls another artifact? This means that an artifact requiring EXECVE can be called from another artifact that doesn't have EXECVE, because it's now called with hard-coded arguments and no longer allows for true arbitrary code execution? Similarly, if an artifact were to pass user-supplied parameter arguments directly to, say, the Windows.System.PowerShell artifact, it should definitely also be specified with EXECE as a required permission.

scudette:
Yes exactly - this is how it works - you can try it by creating a user with limited permissions (no execve):

image

Creating the following artifact
```yaml
name: Custom.PowershellDir
sources:
  - query: |
      SELECT * FROM Artifact.Windows.System.PowerShell(Command="dir C:/")
```

Then try to launch it with that user without any issues:
image

But that user cant launch the original artifact directly.
image


This feature is deliberate - it allows wrapping of dangerous artifacts with preset parameters so low privilege users can use them safely

misje:
Perfect, that is what I was assuming. Thanks for taking the time to explain.

PR #3740, opened 2024-09-09
Remove hard-coded prompt=login parameter from OIDC authenticator

hillu:
We found that if Authentik is configured to refuse authentication using passwords cannot be used with prompt=login. Users only get a message "Flow dows not apply to current user".

After removing the added parameter prompt=login, everything works fine.


hillu:
The extra parameter was added in dd78d2b073b9d98c675d778258888073d0f599a4 – if this change breaks Okta authentication flows, it might make sense to add a config setting.

PR #3985, opened 2025-01-01
Add yara linting example

predictiple:


predictiple:
This example currently produces a non-working rules file because of the non-removal of "import hash" by yara_lint.

PR #4236, opened 2025-05-13
Added LNK file condition check

jwardsmith:
There was no condition check if a file was ending a .lnk extension.

This led to the Artifact.Windows.Forensics.Lnk attempting to parse other files such as executables.


mgreen27:
This is intentional as the parser checks for LNK headers. You should control this via the glob

mgreen27:
My mistake - I see your modifying the startup artifact and not the LNK parser - should be ok.

jwardsmith:
No issues

PR #4317, opened 2025-06-26
Update PowerShell.yaml

randomaccess3:
update default stdout length

### DIFF
--- a/artifacts/definitions/Windows/System/PowerShell.yaml
+++ b/artifacts/definitions/Windows/System/PowerShell.yaml
@@ -40,11 +40,9 @@ parameters:
   - name: PowerShellExe
     default: "C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe"
   - name: StdoutLength
+    type: int
     default: 50000000
 
 sources:
   - query: |
-      SELECT * FROM execve(argv=[PowerShellExe,
-        "-ExecutionPolicy", "Unrestricted", "-encodedCommand",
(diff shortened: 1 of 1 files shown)


scudette:
FYI: This failed because

```
Error with Windows.System.PowerShell: Invalid arg length for function base64encode()
```

randomaccess3:
huh will fix, worked when i tested it but maybe i edited it again

scudette:
Using the latest CI build this error will be shown by the editor now
image

randomaccess3:
Updated the commit, set the length variable to an int and fixed the brackets

scudette:
This is actually not going to work beyond about 3-4mb because the output is used in a JSON row. It will lead to grpc failures which tend to confuse users and break the GUI.

I suggest the output should be uploaded as a file with a preview column type so the user can view it properly.

randomaccess3:
yeah the current process breaks the gui - if you run getaduser it outputs into a lot of rows instead of one. i couldnt get it to parse out the json either way.
I dont mind cutting the output down to under 3mb since most of the powershell stuff should be less than that.

not sure how to do the upload as a file, that might be useful for the get-aduser stuff i was doing though versus modifying the existing powershell execution shell. Although having the option to set the output size to be bigger gives the user some flexibility they didnt have before

scudette:
Since this artifact is actually used by the GUI we need to modify the GUI in concert with it.

scudette:
This should be merged upstream

PR #4457, opened 2025-09-17
Added enable/disable all button #4452

cichyx96:
Hi, it's my first PR, I have tried to implement requested feature #4452.
I have added button to enable/disable all boolean switches.

However there are my observations:
1. Currently it's presented for every artifact, even if it has 0 boolean switches.
2. Not sure how this radio button should act when someone presses ENABLE ALL, and then manually turns off one switch, should it then change state to off? Someone could be confused that set Enable All to true, but in VQL not all booleans ended up as TRUE. It should be clear it's UI feaure and not Artifact Parameter which overrides other parameters.

velociraptor_gif

### DIFF
--- a/gui/velociraptor/src/components/flows/new-collections-parameters.jsx
+++ b/gui/velociraptor/src/components/flows/new-collections-parameters.jsx
@@ -194,12 +194,37 @@ export class NewCollectionConfigParametersForm extends React.Component {
         this.props.setParameters(parameters);
     }
 
-   removeArtifact = (name) => {
+    toggleBooleanParams = (artifact, on) => {
+        let parameters = this.props.parameters;
+        let artifact_parameters = (parameters[artifact] || {});
+
+        // Find the artifact descriptor to know param types
+        const descriptor = _.find(this.props.artifacts, x=>x.name === artifact);
+        const boolParamNames = _.map(
+            _.filter((descriptor && descriptor.parameters) || [], p=>p.type === 'bool'),
+            p=>p.name
+        );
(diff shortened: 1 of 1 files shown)


scudette:
Thanks for this . Do you think it would be too cluttered if the button appears as a flag itself on every form? perhaps it is better to have that as a context menu (right click) so it does not take real estate since it is rarely used?

cichyx96:
I think having it in context menu isn't very intuitive (I doubt people would find it out wihtout documentation or by accident), I can try to code it so it only appears when there are 2 or more booleans, also Radio button might be not best choice for this functionality

scudette:
I guess the question is - what problem are we trying to solve? This is an issue in artifacts like KapeFiles.Targets that have a lot of bool values but usually the bool values are rarely all on or all off.

Maybe it makes more sense to refactor artifacts like KapeFiles to have a better UI element that serves the purpose better? like maybe a multi-select would be better?

This way we can keep the targets separate from flags like `UPLOAD_IS_RESUMABLE` which is functionally unrelated to the targets. Using a global all on or all off button will lump the `UPLOAD_IS_RESUMABLE` options together with the targets and make it harder to use actually.

scudette:
I have refactored the KapeFiles artifact to use a multichoice instead and I think it looks a lot better

image


The new interface is split into higher level targets which there are only 5-6 of and all the specific low level targets are below. 

As such I dont think this PR is needed. The targets are separated from the other boolean flags which are not logically related anyway.

PR #4672, opened 2026-02-21
fix(reformat): reformat overwrites content

nullifysecurity:
This pull request fixes an issue where the current reformat logic could unintentionally overwrite content following the `query` field in an artifact. This most commonly affected artifacts that define additional fields such as `notebook` or `column_types`.

For example, given the following artifact with a `notebook` field:

```yaml
name: Notebook
sources:
- query: |
    SELECT A,B,C
    FROM scope()
  notebook:
    - name: Test
      type: vql_suggestion
      template: |
        SELECT * FROM scope()
```

Running the reformat command (either via CLI or the GUI) would remove the `notebook:` line. Subsequent reformats would continue stripping content, producing an invalid artifact such as:

```yaml
name: Notebook
sources:
- query: |
    SELECT A,
           B,
           C
    FROM scope()
    - name: Test
      type: vql_suggestion
      template: |
        SELECT * FROM scope()
```

> Error: While parsing VQL at line 8: unexpected token "-"

The root cause was an off-by-one error in the mutation logic applied to the `query` block, which caused the reformat operation to extend beyond the intended range. While addressing this, I also identified an issue with how `original_end_line` was calculated for chomping indicator nodes (`|-`). Because these nodes omit the trailing newline, splitting on `\n` resulted in an incorrect end line calculation.

Additionally, the `current_mu.err` value was never checked when applying mutations that only had a length of one. This has also been resolved by adding an additional check at the beginning of the line iteration.

This change corrects these issues and adds regression tests to verify that:
- Content following a `query` block is preserved.
- Running reformat multiple times produces stable, idempotent output.

Please let me know if you'd like further clarification, or if there are any changes you'd like made. 

I am also considering a refactor of the `artifacts reformat` to call a `reformat` VQL function which wraps this functionality, following the pattern that CLI commands are just wrappers around VQL. Let me know if this is also something you'd be interested in me contributing. 😃

### DIFF
--- a/services/repository/fixtures/TestReformat.golden
+++ b/services/repository/fixtures/TestReformat.golden
@@ -23,14 +23,16 @@
   "- query: SELECT * FROM info()",
   "- query: |",
   "    SELECT A",
-  "    FROM scope()"
+  "    FROM scope()",
+  ""
  ],
  "export reformatted": [
   "",
   "name: Exported",
   "export: |",
   "  SELECT *",
-  "  FROM scope()"
+  "  FROM scope()",
+  ""
  ],
  "preconditions": [
   "",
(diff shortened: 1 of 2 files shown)


scudette:
Thanks for this! +1 on the reformat() VQL function idea - it would be very helpful.

PR #4691, opened 2026-02-27
Added more detailed logging for VFS Downloads

AmazingStarBottle:
I added a quick logger.Debug to ProcessDownloadFile, to log the user that downloaded a file, the path of the downloaded file, the client ID and the flow ID.

This will make it clearer from the logs, which user performed the action, without having to check in the GUI

### DIFF
--- a/services/vfs_service/vfs_service.go
+++ b/services/vfs_service/vfs_service.go
@@ -125,6 +125,20 @@ func (self *VFSService) ProcessDownloadFile(
 		SHA256, _ := row.GetString("Sha256")
 		Error, _ := row.GetString("Error")
 
+		// Code to log metadata about the downloaded file.
+		user := ""
+		if flow.Request != nil {
+			user = flow.Request.Creator
+		}
+
+		logger.Debug(
+			"VFSService: Collecting %s from %s for %s, flow=%s",
+			Path,
+			client_id,
+			user,
+			flow_id
+		)
+
 		// Figure out where the file was uploaded to.
 		uploaded_file_manager := flow_path_manager.GetUploadsFile(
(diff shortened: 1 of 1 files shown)


AmazingStarBottle:
Not sure if it should be a Debug-level or Info-level

scudette:
By default debug level log is disabled. 

If you want to know about who scheduled the file you can easily write a server monitoring artifact.

See this https://docs.velociraptor.app/artifact_references/pages/system.flow.completion/

AmazingStarBottle:
Could it be possible to still add it, just as a debug-level log? Or is it a preference for you guys, to use artifacts when possible?

scudette:
I am just trying to understand what problem you are trying to solve with this?

Is it a matter of auditing? Do you want to know if user X downloaded file Y? This information is in the audit log

The log messages in this PR will only trigger if the file was successfully fetched when it is fetched - while the audit log says when the user initiated the collection request (not necessarily if it completed or not).

Usually debug messages are turned off in production so this PR wont help it. We can add it to info level but then the question is "how would you use this message?". It likely requires a pipeline to filter it out of the normal stream of logs.

AmazingStarBottle:
When I looked at the logs that were generated when a user used the VFS feature to download a file, I could not see it anywhere in the syslog logs. I could see that the DownloadFile feature was used, as well as a flow ID and a client ID, but no name of the file, or even the user that did it. This was with all logs being forwarded, all logs enabled.

That is why i made this change, to include important information like the username of the user performing the action

scudette:
You are 100% correct, thanks for reporting it. I just verified and there was a missing audit message on this action.

This should be fixed in #4692

PR #4702, opened 2026-03-03
Strip wrapping single quotes from file path parameters

### LINKED ISSUE #4701: Strip wrapping single quotes from pasted file path parameters
On macOS 15+, Finder's "Copy as Pathname" wraps paths containing spaces in single quotes (e.g. `'/Users/admin/Desktop/my file.json'`). When pasted into artifact parameter fields in the GUI, the literal quote characters are included in the file path, causing silent failures and returning 0 rows.

araesa:
Fixes #4701

On macOS 15+, Finder's "Copy as Pathname" wraps paths containing spaces in single quotes (e.g. `'/Users/admin/Desktop/my file.json'`). When pasted into artifact parameter fields in the GUI, the literal quote characters are included in the file path, causing silent failures and returning 0 rows.


scudette:
I think this is a bug in MacOS which is pretty widely discussed - there are some [workarounds](https://tinyapps.org/blog/copied-pathname-quotes.html) so there is no point in changing our code to fix a MacOS bug.

PR #4708, opened 2026-03-06
Update FSEvents.yaml

herpf:
Hi team,

I was doing a hunt recently and noticed this artifact wasn't returning folders I knew existed in the FSEvents logs. I dug into the VQL and found two issues that this PR fixes:

1. Filtering logic bug The `WHERE EntryPath =~ ...` filter was placed inside the `LET x` assignment. Since `EntryPath` and `EntryFlags` are only aliased in the final `SELECT` statement, they evaluated to Null during the `LET` phase, dropping valid data before it could be flattened. I moved the `WHERE` clause to the very end of the query.

2. File truncation: In the `parse_binary` function, `read_file` was hardcoded to a 1MB limit (`length=1000000`). Uncompressed FSEvents logs often exceed 1MB, which silently truncates the data and misses events at the tail end of the file. I increased this limit up to 10MB (`length=10000000`).

I tested the updated VQL locally and it successfully catches the missing events now. Let me know if you need any changes!

### DIFF
--- a/vql/windows/etw/context.go
+++ b/vql/windows/etw/context.go
@@ -101,7 +101,13 @@ func (self *Handle) Close() {
 	self.closed = true
 
 	self.cancel()
-	close(self.output_chan)
+	select {
+	case <-self.ctx.Done():
+		return
+
+	default:
+		close(self.output_chan)
+	}
 }
 
 // Try to send but skip closed handles.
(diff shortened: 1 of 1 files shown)


artifacts/definitions/MacOS/Forensics/FSEvents.yaml
```diff
@@ -232,11 +232,10 @@ sources:
                 Btime as SourceBtime
             FROM
                 foreach(row=parse_binary(
-                    filename=read_file(filename=OSPath, accessor="gzip", length=1000000),
+                    filename=read_file(filename=OSPath, accessor="gzip", length=10000000),
```
scudette: Maybe make this into a parameter with default 10m? Do you think files may exceed this size?

herpf: In my lab endpoint which I heavily use I see the largest single file being 0.25 MB after decompressing but we can change it to a parameter just to be on the safe side.

PR #4721, opened 2026-03-11
Improve Debian Packages artifact

misje:
I am using this artifact a lot, but only recently I discovered the discrepancy between this list and the list of actually installed packages. Previously installed packages, either removed or purged, will still be accounted for by dpkg. This may or may not be what the investigator was looking for.

Currently, this artifact will list deb packages and snaps that were once installed, but has since been removed. This is misleading, in my opinion. Still, gathering the traces of these installs are very useful. I propose these changes:

- Instead of a singular combined notebook cell with a combined result,
  - Provide all raw results from debs (now including a state/status field)
  - Provide all raw results from snaps (now including a state/status field)
  - Continue providing a combined table, but focus on just installed apps

This also fixes bugs that would previously

  - Assume that an app is installed, but it was only partially installed, broken or removed


misje:
I forgot to add the new columns to the tests, sorry about that. Thank you for doing that!

scudette:
Not a problem! Thanks for the PR

misje:
I realised that I made a silly mistake. Fixed in #4725.

PR #4809, opened 2026-04-28
New flow upload artefact for azure data explorer (adx)

Che4ter:
This pull requests adds a new server_event artefact to forward flow results to Azure Data Explorer (ADX).
ADX is the data analytics platform from microsoft azure, which allows the analysis of the data using kql queries.
It mirrors the existing artefact for uploading flows like the splunk_upload and the elastic_upload.
We've tested the new artefact already during several smaller engagement and so far it works fine.


scudette:
Thanks for this - I checked and it adds about 3mb to the binary size. Should we restrict it to the sumo build? I imagine it is only useful on a server?

Che4ter:
@scudette Yes its a server_event artifact so no need to include it in the client.
I've added
//go:build sumo
// +build sumo
to the adx upload file. Is that the correct approach?

artifacts/definitions/ADX/Flows/Upload.yaml
```diff
@@ -0,0 +1,137 @@
...
+               WHERE Flow.artifacts_with_results =~ ArtifactNameRegex
+      LET organization <= org().name
+
+      LET documents = SELECT * FROM foreach(row=completions,
+          query={
+             SELECT * FROM foreach(
+                 row=Flow.artifacts_with_results,
+                 query={
+                     SELECT *, _value AS Artifact,
+                            client_info(client_id=ClientId).os_info.hostname AS Hostname,
```
scudette: If these fields are interpreted by the plugin we need to make sure they dont get mixed up with whatever columns the artifacts emit- maybe prefix them with _ like _ts, _hostname etc.

services/secrets/initialize.go
```diff
@@ -112,6 +112,22 @@ var (
 			"addresses":   "# Add URLs one per line\n# http://www.example.com/\n",
 			"skip_verify": "FALSE",
 		},
+	}, {
+		TypeName:    constants.ADX_CREDS,
+		Description: "Credentials to be used in adx_upload() calls.",
+		Verifier:    "x=>x.client_id AND x.client_secret",
```
scudette: I guess there are more fields that must be set right? like tenant id?

vql/server/adx.go
```diff
@@ -0,0 +1,509 @@
...
+	batchIngestionTime := ""
+	var batchStart time.Time
+
+	// Create connection string with service principal auth
+	kustoConnectionString := azkustodata.NewConnectionStringBuilder(arg.ClusterURL).
+		WithAadAppKey(arg.ClientID, arg.ClientSecret, arg.TenantID)
+
+	// Azure Kusto ingest-* endpoints (e.g. ingest-*.kusto.windows.net) do not
+	// support TLS 1.3, therefore force TLS 1.2 for now
+	httpClient := &http.Client{
```
scudette: it is probably useful to use the generic http client from https://github.com/Velocidex/velociraptor/blob/b390809e6dfa81fd0f8f691b71f62b67597de926/vql/networking/transport.go#L65 so we support proxies, etc. You can overwrite the tls config from there if you need to

vql/server/adx.go
```diff
@@ -0,0 +1,509 @@
...
+	ctx context.Context,
+	scope vfilter.Scope,
+	config_obj *config_proto.ClientConfig,
+	output_chan chan vfilter.Row,
+	row_chan <-chan vfilter.Row,
+	wg *sync.WaitGroup,
+	arg *_ADXPluginArgs) {
+	defer wg.Done()
+
+	var buf = make([]vfilter.Row, 0, arg.ChunkSize)
```
scudette: it is generally better to serialize the row ASAP and just keep the jsonl around in the chunk:
1. It takes a lot less memory 
2. It forces the row to materialize thereby destroying any references to long lived objects so the memory can be collected immediately.

Once the chunk is full, the jsonl can be sent at once.

vql/server/adx.go
```diff
@@ -0,0 +1,509 @@
...
+	opts *json.EncOpts) error {
+
+	// Check context cancellation before expensive operations
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	default:
+	}
+
+	vql_row_dict := vfilter.RowToDict(ctx, scope, row)
```
scudette, line 313: This is already a new dict so you dont need to copy it again below.

vql/server/adx.go
```diff
@@ -0,0 +1,509 @@
...
+	// Create RawData by copying vql_row_dict and deleting metadata fields (optimized)
+	rawData := vql_row_dict.Copy()
+	rawData.Delete("Artifact")
+	rawData.Delete("ClientId")
+	rawData.Delete("FlowId")
+	rawData.Delete("Organization")
+	rawData.Delete("Hostname")
+
+	// Create structured row with metadata + RawData
+	structuredRow := ordereddict.NewDict().
```
scudette: I suggest to use json.Format to get better performance with less allocations.

https://github.com/Velocidex/velociraptor/blob/b390809e6dfa81fd0f8f691b71f62b67597de926/json/shortcuts.go#L40

artifacts/definitions/ADX/Flows/Upload.yaml
Che4ter: I prefixed them with _

services/secrets/initialize.go
Che4ter: I've added the tenant_id as well. But the other fields are not validated because we have usually one secret configurated in the root org that holds the tentant specific settings. Then we use different orgs for our engagements that depending on the type have their own adx cluster or database. For those we override the specific settings in the individual artifact configs.
I hope this makes sense?

vql/server/adx.go
Che4ter: Good to know, I changed it.

vql/server/adx.go
Che4ter: thanks I removed the copy

vql/server/adx.go
Che4ter: I removed the vfilter.Row and serialize it directly now.

PR #4849, opened 2026-05-26
Spelling fixes

predictiple:
that Vale didn't catch these last time.


scudette:
Can you also please add a make target to run vale on the current repo manually?

predictiple:
Sure but I'll need to move all the ancilliary files into the repo, which is just over 3MB.

Screenshot from 2026-05-26 04-25-30

The convention recommended by Vale is to put them in the .github directory, but if you prefer we can put them in `.vale` in the root?

scudette:
Are these files likely to change often?

predictiple:
Yes I tweak them all the time. It's a WIP. I created most of those hunspell dictionaries and update them rather than using exclusion lists.

But it's 1 command to run all the checks and that stays constant.

predictiple:
I can add it and you can try it and if you don't like it you can delete it in your next PR.

scudette:
I guess I am asking if these are binary files that change frequently they will lead to a lot of git churn- if they are text files that change frequently it is not an issue unless a lot of the file changes at once. 

If they are still work in progress and are unsuitable to be included in the tree, we should commit them into a separate Repository then either add it as a git submodule or have the makefile launch a script which checks these out on demand into a subdirectory.

predictiple:
Oh I see. They are all text files.

Change frequency is low. For example, for this PR I added some terms to the infosec dictionary.

Importantly they are applicable to all repos (i.e. to ensure consistency). Each repo has it's own config file specifying what to check but the checks themselves and the supporting files all live in this 1 directory tree. So it would be better to have them located centrally so that improvements automatically apply to all.

predictiple:
I wrote this for the docs but then pulled it. This is how it would work if done the conventional way.

---

## What is Vale?

https://vale.sh/docs

> Vale is a command-line tool that brings code-like linting to prose.
> Vale is cross-platform (Windows, macOS, and Linux), written in Go,
> and available on GitHub.

## Vale files and folders

By cloning the repo you will already have the files and folders
necessary to run the Vale tests. All you need to do is ensure that you
have the Vale binary in your path on your computer so that you can
actually run the tests.

`.vale.ini` is the key file, located in the project root.

This file defines which file targets to lint, what linting rules to
apply, and the overall alert threshold that should be reported.

The `styles` **folder**, which in `.github/` in the project root,
contains the linting rules (each defined as YAML - much like
Velociraptor artifacts) plus support files such as dictionaries and
"views" definitions that the rules need to operate. The folder
structure we use is
[as recommended by Vale](https://github.com/vale-cli/vale-action#repository-structure).


## Running Vale locally

1. Install the latest `vale` binary into your path. Test it by running
   `vale --ls-dirs` or `vale version`.

   NOTE: despite what their documentation says you DO NOT need to run
   `vale sync`. We don't use external styles and therefore don't need
   anything outside of our own repo. Everything you need, except for
   the Vale binary, is already in the repo.

2. `cd` to the project root - when you run Vale it will look for the
   `.vale.ini` file in the `CWD`, which in turn will tell Vale where
   to find the Vale `styles` directory (which is in `.github/`).

3. Run `vale .` is all you need to do to perform all the checks.

PR #4857, opened 2026-06-01
Windows.Forensics.RDPCache: apply DateAfter/DateBefore to Parsed source

### LINKED ISSUE #3781: RDPCache artifact time filter
We need to add modified time filter to RDPcache artifact to help targeting files in scope faster. Current workflow need to run, use notebook then rerun targeting machines. Would be much easier to just add the filter on collection.

cray44:
Fixes #3781.

The DateAfter/DateBefore parameters were only applied in the TargetFiles source. This applies them to the Parsed source as well, so .BIN files outside the requested window are filtered out before parse_binary() runs instead of being parsed and discarded in a notebook afterward.

The filter uses the Mtime of the .BIN container file. Individual cached bitmaps carry no timestamps, so the container Mtime is the finest granularity available. Boundaries are exclusive, matching TargetFiles.

### DIFF
--- a/artifacts/definitions/Windows/Forensics/RDPCache.yaml
+++ b/artifacts/definitions/Windows/Forensics/RDPCache.yaml
@@ -70,6 +70,12 @@ sources:
   - name: Parsed
     description: Parsed RDP BitmapCache files.
     query: |
+      -- Scope parsing to the requested DateAfter/DateBefore window.
+      LET DateAfterTime <= if(condition=DateAfter,
+        then=DateAfter, else=timestamp(epoch="1600-01-01"))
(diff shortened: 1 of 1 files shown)


cray44:
@mgreen27 this implements the Parsed-source time filter you asked for in #3781. The `DateAfter`/`DateBefore` bounds are now applied in the `Parsed` source the same way they already were in `TargetFiles`, so out-of-window `.BIN` files are skipped before `parse_binary()` runs. Filtering is on the container `Mtime` (individual cached bitmaps have no timestamps). Happy to adjust if you'd prefer a different approach.

PR #4872, opened 2026-06-09
Refactored metadata to avoid get/modify/set pattern

scudette:


misje:
The binary runs with `gui`, but when I tried it out in a production environment (with plenty of clients, and probably client metadata updating in the background), various things did not work. For instance the client list and server collection view did not load. Nothing in the log. It feels like there is a deadlock.

misje:
I am testing your new changes. So far, so good. Just to be clear, as long as I do not depend on other metadata when updating metadata, I should be fine just setting a new dict in `metadata` with the fields I need? I do not need to use the new callback feature.

PR #4873, opened 2026-06-09
Add timeline VQL suggestions, and make journal/evtx suggestions similar

misje:
- Add more useful columns to simplified journal view
- Add message count plot also to journal
- Add a timeline VQL suggestion to EVTX
- Add a timeline VQL suggestion to journal

### DIFF
--- a/artifacts/definitions/Linux/Forensics/Journal.yaml
+++ b/artifacts/definitions/Linux/Forensics/Journal.yaml
@@ -26,6 +26,16 @@ parameters:
   type: bool
   description: If set we also upload the raw files.
 
+export: |
+  LET Priorities <= dict(`0`='emerg',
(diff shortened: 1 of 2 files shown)


artifacts/definitions/Linux/Forensics/Journal.yaml
```diff
@@ -105,17 +105,77 @@ sources:
...
   notebook:
     - type: vql_suggestion
-      name: Simplified syslog-like view
+      name: Simplified view
       template: |
         /*
         # Simplified log view
         */
-        LET ColumnTypes<=dict(`_ClientId`='client')
+        LET Priorities <= dict(`0`='emerg',
```
scudette: Adding this to the export section should make it execute before any cell in the notebook.

PR #4885, opened 2026-06-16
linux: fall back to direct /proc read for pslist(pid=X) on rootkit hidden processes

### LINKED ISSUE #4761: linux: pslist(pid=X) on "hidden" processes does not work
Some LKM rootkits and BPF-based [boopkit](https://github.com/krisnova/boopkit) hide processes by hiding `/proc/$PID` from the output produced by `getdents64(2)`. In all cases known to me, the hidden directory can still be traversed. In these cases, pslist should just produce the fields it would normally produce. 

Trying to dump process memory for hidden processes using the `Linux.Triage.ProcessMemory` artifact does not work right now even though `/proc/$PID/{maps,mem}` are available.

As far as I can see, changing `pslist()` behavior would require changing `gopsutils/process.PidExistsWithContext`.

litemars:
## Summary
- Fixes #4761: `pslist(pid=X)` on Linux fails for processes hidden by LKM/eBPF rootkits that intercept gopsutil's existence probe, even though `/proc/<pid>/...` remains directly traversable.
- Adds a Linux-only direct `/proc` reader as a fallback when gopsutil's `process.NewProcessWithContext` errors.
- Normal visible-process behavior is unchanged: the fallback only runs on the error branch.

## Root cause
`GetProcess()` currently relies on `process.NewProcessWithContext()` for targeted PID lookups and the current Linux implementation in Velociraptor does not have a fallback path when that check fails.

In the scenario described in the issue above, some rootkits hide a process from the usual `/proc`-based existence checks, while direct access to `/proc/<pid>/status`, `/proc/<pid>/maps` or `/proc/<pid>/mem` can still succeed. As a result, `pslist(pid=X)` reports the process as missing even though targeted `/proc` access is still possible.

### DIFF
--- a/vql/psutils/process_posix.go
+++ b/vql/psutils/process_posix.go
@@ -17,6 +17,9 @@ import (
 func GetProcess(ctx context.Context, pid int32) (*ordereddict.Dict, error) {
 	process_obj, err := process.NewProcessWithContext(ctx, pid)
 	if err != nil {
+		if direct, directErr := GetProcessDirect(ctx, pid); directErr == nil {
+			return direct, nil
+		}
 		return nil, err
 	}
 
--- a/vql/psutils/process_posix.go
+++ b/vql/psutils/process_posix.go
@@ -17,10 +17,26 @@ import (
 func GetProcess(ctx context.Context, pid int32) (*ordereddict.Dict, error) {
 	process_obj, err := process.NewProcessWithContext(ctx, pid)
 	if err != nil {
-		if direct, directErr := GetProcessDirect(ctx, pid); directErr == nil {
-			return direct, nil
+		p := &process.Process{
+			Pid: pid,
 		}
-		return nil, err
+		p.CreateTimeWithContext(ctx)
+
+		// Check if the process is really there.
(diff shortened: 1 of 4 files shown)


vql/psutils/process_posix.go
```diff
@@ -17,7 +17,26 @@ import (
 func GetProcess(ctx context.Context, pid int32) (*ordereddict.Dict, error) {
 	process_obj, err := process.NewProcessWithContext(ctx, pid)
 	if err != nil {
-		return nil, err
+		p := &process.Process{
```
scudette, line 20: Looking at https://github.com/shirou/gopsutil/blob/d781152c7fcf353e319f19bba54147e58de8990b/process/process.go#L217

looks like this can fail the PidExistsWithContext check but the rest of the function looks like it can not fail so we can replicate it here to fallback avoiding the check.

Can you please test this on the rootkit you were looking at? Avoiding reimplementing all the logic again will help make the code more maintainable in future.

scudette:
I incorporated it into #4892  so closing.

PR #4907, opened 2026-06-25
Linux writeback path correction in sample config

predictiple:


scudette:
That's actually not a good default location. The etc directory is not writable for non root and tmp is guaranteed to exist and be writable so that's why it's the default. In practice of course it will always be changed to fit with the package so it doesn't matter but at least for velociraptor GUI command that's a good default

predictiple:
Agreed, but it's nevertheless still the default. So this just makes the documentation match reality. Deployments that went with the defaults - probably most, I'd say - will be using that location. 

`/tmp` is not ideal either because of it's non-persistence on most modern Linux distros. Probably the most appropriate location would be `/var/lib/velociraptor/`.

scudette:
It's only the default when the user only runs config generate. The wizard doesn't use that and it's expected to be set to something

The default is not really used for real deployments either
 It can be set to change me to make sure people change it but it's just a starting point and people are expected to modify it

predictiple:
`config generate` with or without the wizard both set it to `/etc`. As far as I can tell the only time it's not the default is when using gui (which will also prefer to use the datastore path if it's specified) ~~or pool_client~~ which use `/tmp`.

Unless there's been some advice presented to users that the default is not suitable for real deployments, or a suggestion of a better/preferred/recommended location, we can assume that this is how most deployments are. If the path was customized then it wouldn't matter to those users what the defaults are because they're not using them. It matters to users who _didn't_ customize it and might refer to the docs to determine where the file is by default.

This is just about making the documentation correct. Either it is or it isn't, and to me it looks like it isn't.

Screenshot from 2026-06-26 13-28-38

scudette:
Maybe this is the real issue, what is the default. For this setting there is actually no default because it has to be set to something. There are many such settings and I think the sample config file specifies something there but it's interpreted as the default. 

Perhaps the docs needs to distinguish between a default and a mandatory setting that should be set. There is no default and I think the config will refuse to load if the setting is empty.

predictiple:
Yes, the description is subtle in that it doesn't actually say that those values shown are default values.

We could say that these are "illustrative values", but that raises the questions "what should it be set to?"  and "what is it usually set to?". For something as fundamental as the writeback file we need to be quite clear about it. We can't be saying "it's whatever you set it to, and it depends how you created your config, and we can't actually tell you what it should be, and the value it has by default is not recommended for production deployments".

If we want to be clear that we're showing a placeholder, we could use something like: `/path/to/your/velociraptor.writeback.yaml`. But that raises the questions I mentioned above.

It's reasonable for users to ask what any setting will be if they didn't change it, and that's interpreted by most people as being defaults. If they generated their config in a recommended way then it's fair to interpret the values produced as defaults. If values produced are not recommended then they shouldn't be the defaults (it should rather produce empty values, or force a decision in some other way).

PR #4911, opened 2026-06-29
Correct writeback_linux default path from /tmp/ to /etc/

sedoy107:
Fixing a discrepancy in the documentation that shows the default path for writeback file on linux as `/tmp/velociraptor.writeback.yaml` while it is actually `/etc/velociraptor.writeback.yaml`.

Also, I went ahead and updated the test fixture files with the correct value.


sedoy107:
As per the comment in Discord, reverted the previous commit that touched the tests and updated only the documentation reference

PR #4924, opened 2026-07-16
fix: update artifact default timeout behaviour

nullifysecurity:
This pull request makes a minor change to the artifact timeout handling which brings its behaviour inline with user expectation that defining a timeout on a collection overrides any default timeouts defined on individual artifacts.

Currently, if a default timeout is specified on an artifact through its definition (such as below), any changes to the timeout specified by the user in the collection are not reflected in the request sent to the client.

```
resources:
  timeout: 3600
```

This leads to unexpected behaviour such as the query timing out after the timeout specified on the collection, or, if a longer timeout is specified on the collection, the query will timeout before the collection times out, shown below.

image
image

The [resources documentation](https://docs.velociraptor.app/docs/artifacts/advanced_fields/#-resources-) describes these as "defaults", but as a user, I would expect that defining a timeout on the collection would override any default timeouts that are set on individual artifacts.

I can see the use case for defining these default timeouts for individual artifacts, but maybe there's a way that we can support both behaviours? Let me know your thoughts!

### DIFF
--- a/services/launcher/launcher.go
+++ b/services/launcher/launcher.go
@@ -336,8 +336,8 @@ func (self *Launcher) CompileCollectorArgs(
 				vql_collector_args.IopsLimit = collector_request.IopsLimit
 			}
 
-			if vql_collector_args.Timeout == 0 &&
-				collector_request.Timeout > 0 {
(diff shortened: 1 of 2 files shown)


scudette:
Thanks, this makes sense!

PR #4925, opened 2026-07-17
fix: refresh cached tool hashes

nullifysecurity:
This pull request addresses an edge case that we encountered recently where a hunt was scheduled with an artifact containing a tool definition, but the tool definition was changed before the collection was scheduled on the host. This led to a hash mismatch when the tool was fetched on a client and subsequently caused the collection to fail.

My proposed solution to this scenario is to refresh the tool dependencies when a collection is scheduled and the request contains pre-compiled collector args. Of course this comes with a slight overhead of having to query the inventory and update the request, but hopefully this shouldn't slow down the scheduling process too much.

I have also added an `upsertVqlEnv` helper function to prevent code duplication when refreshing the tool dependencies in `ScheduleArtifactCollection` and to ensure that the `VQLEnv` is only added once.

Let me know if I've overlooked anything or this causes any unintended side-effects.

### DIFF
--- a/services/launcher/launcher.go
+++ b/services/launcher/launcher.go
@@ -534,18 +534,6 @@ func (self *Launcher) EnsureToolsDeclared(
 	return nil
 }
 
-// Update an existing VQLEnv if it exists, otherwise append it.
-func upsertVqlEnv(existing []*actions_proto.VQLEnv, env *actions_proto.VQLEnv) []*actions_proto.VQLEnv {
-	for idx, item := range existing {
-		if item.Key == env.Key {
-			existing[idx] = env
-			return existing
-		}
-	}
-
-	return append(existing, env)
-}
-
 func AddToolDependency(
 	ctx context.Context,
 	config_obj *config_proto.Config,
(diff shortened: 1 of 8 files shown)


scudette:
When the hunt is created the artifact is compiled into raw VQL statements and it is at this point that the tools are expanded. 

After the hunt is created the hunt manager just uses the same pre-compiled request for each new client. So it no longer depends on the tool definition.

The request contains the URL from where the tool will be downloaded and that depends only on the filename (and org) of the tool.  Are you saying the binary was changed on the server while keeping the same name? How was the tool definition changed ?

services/launcher/launcher.go
```diff
@@ -623,6 +635,21 @@ func (self *Launcher) ScheduleArtifactCollection(
...
 		args = append(args, compiled...)
+	} else {
+		// Collector args are cached for performance but may contain stale tool information.
+		// This ensures that before an artifact is scheduled, it has the most up-to-date tool data.
+		for idx, arg := range args {
+			_artifacts := arg.GetArtifacts()
+
+			for _, artifact := range _artifacts {
+				for _, tool := range artifact.Tools {
+					err := AddToolDependency(ctx, config_obj, tool.Name, tool.Version, args[idx])
```
scudette: Doing this for each scheduled client is a performance killer. The compiled request should be reused exactly.

nullifysecurity: This was my main concern with this solution, thanks for confirming.

nullifysecurity:
> The request contains the URL from where the tool will be downloaded and that depends only on the filename (and org) of the tool. Are you saying the binary was changed on the server while keeping the same name? How was the tool definition changed ?

In this case, yes, the binary was changed on the server while keeping the same name. We use Velociraptor's tool system to deliver up-to-date detections and indicators to artifacts that depend on them and utilize a server artifact that reaches out to an endpoint every hour to download the latest definitions. If there has been an update to our definitions, the server artifact calls `inventory_add` to update the tool definition with these changes. Understandably, this not an intended use-case for tools to be updated  so frequently, hence why it's likely an edge-case in the way that Velociraptor handles the retrieval of tools.

Here's what I was able to gather from troubleshooting this issue and coming up with my proposed solution (which I note due to your comment is very bad for performance and confirms my original concerns).

When the hunt is created and the artifact is compiled and stored in `collector_request.CompiledCollectorArgs`, the tool definition is defined in scope using the following variables:

- `Tool_ExampleTool_HASH`
- `Tool_ExampleTool_FILENAME`
- `Tool_ExampleTool_PATH`
- `Tool_ExampleTool_URL`
- `Tool_ExampleTool_URLs`

When `Generic.Utils.FetchBinary` is called on the client to retrieve the tool, it reads these variables from the scope, when determining where to download the tool from, and what its expected hash is. If the definition of the tool changes in the time between hunt creation and scheduling on the client, either through manual overrides or using `inventory_add`, the variables defined in the scope become stale and cause `Generic.Utils.FetchBinary` to fail on the hash validation,

```
LET args <= dict(
  ToolHash=get(field="Tool_" + ToolName + "_HASH"),
  ToolFilename=get(field="Tool_" + ToolName + "_FILENAME"),
  ToolURL=get(field="Tool_" + ToolName + "_URL"),
  ToolURLs=ParseUrls(Url=get(field="Tool_" + ToolName + "_URLs")))

...

WHERE log(message=format(format="downloaded hash of %v: %v, expected %v", args=[
  Content, Hash.SHA256, args.ToolHash]))
    AND Hash.SHA256 = args.ToolHash

```

Maybe a better solution would be to solve this problem at the artifact level by modifying the `Generic.Utils.FetchBinary` artifact to take a `VerifyHash` parameter that would allow deviations between the fetched file hash and the expected hash; although we then lose the safety of making sure that the file we retrieved wasn't modified in transit or has been replaced with a malicious version.

scudette:
Would it make more sense to use the tool's version and tool name in the ultimate path:

https://github.com/Velocidex/velociraptor/blob/87b9ec3bcbd4ac0e24f1807b84b06e03e5ef207d/services/inventory/inventory.go#L535

Ideally each tool + version should be absolutely unique and never change. This is the core assumption with semantic versioning. So if the uploaded files for each version are never overwritten then there should be no confusion.

nullifysecurity:
That definitely makes a lot more sense, thanks Mike!

I'll quickly implement this and add a new commit to this pull request.

nullifysecurity:
I've also added `Tool_%v_VERSION` to `vql_collector_args.Env` as a quality of life feature in case consumers want to be able to perform logic based on the current tool version in scope from their artifacts.

PR #4933, opened 2026-07-20
Add Journal.Fields artifact

misje:
This replaces velocidex/velociraptor-docs#1287. The included journal files do not provide a lot of test data, but at least some.


scudette:
Do you want it upstream or in the exchange? we can not reuse the same name in both as the exchange will refuse to load a built in artifact

PR #4950, opened 2026-08-02
Upgrade to participle v2

predictiple:
Use github.com/alecthomas/participle/v2 v2.1.4 directly instead of v0.7.1. This is required to consume the v2-based vfilter branch which adds source position capture to the VQL AST (for future LSP support).

- vql/windows/wmi/parse: convert lexer from a single mega-regex with named groups to lexer.SimpleRule list, use generic MustBuild, and adapt to the new ParseString signature.
- services/repository/errors.go: UnexpectedTokenError is now returned as a pointer in v2.
- services/repository/reformat.go: participle.Error interface now exposes Position() instead of Token().

Enable the local replace for vfilter while prototyping. participle v0.7.1 remains in the graph as an indirect dependency of sigma-go.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,7 +19,7 @@ require (
 	github.com/Velocidex/zip v0.0.0-20251027040802-582e676739bd
 	github.com/alecthomas/assert v1.0.0
 	github.com/alecthomas/chroma v0.7.3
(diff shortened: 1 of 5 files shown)


predictiple:
Superseded by https://github.com/Velocidex/velociraptor/pull/4956

PR #4960, opened 2026-08-06
fix: ArpCache - no additional encoding needed for State

NicknameNotTaken:
This Artifact fails in with its current definition. Per the documentation, the function "encode" does not support any type "string". 
Since the State is always as uint8, the function does not seem necessary. 

Tested on Win11.

### DIFF
--- a/artifacts/definitions/Windows/Network/ArpCache.yaml
+++ b/artifacts/definitions/Windows/Network/ArpCache.yaml
@@ -45,8 +45,8 @@ sources:
                     then="Active",
                   else="?")) as Store,
 
-               get(item=parse_json(data=kMapOfState),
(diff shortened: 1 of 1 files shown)


artifacts/definitions/Windows/Network/ArpCache.yaml
```diff
@@ -45,8 +45,8 @@ sources:
                     then="Active",
                   else="?")) as Store,
 
-               get(item=parse_json(data=kMapOfState),
-                   member=encode(string=State, type='string')) AS State,
+               get(item=parse_json(data=kMapOfState), member=State) AS State,
```
scudette: Maybe use `str(str=State)` - this code is suppose to convert the state to a string because dicts always have a string key but State may be an integer.

PR #4964, opened 2026-08-07
readers: hold the lock while reading through a pooled AccessorReader

### LINKED ISSUE #4963: Silent evidence corruption: AccessorReader.ReadAt reads through a pooled reader after unlocking, so a concurrent Close() can land mid-read
### Summary

`AccessorReader.ReadAt` in `vql/readers/paged.go` releases its mutex **before** reading
through `self.paged_reader`. Because that reader lives in a scope-wide pool that can close it
concurrently, a read can land on a handle that is being closed, and the read fails with
`os.ErrClosed`.

Callers that treat a read error as "no data" rather than "retry" then emit **wrong evidence,
silently** — no error, no log line, and an unchanged row count. `authenticode()` is the worst
case: it returns `SubjectName`/`IssuerName` as null and `Trusted` as the literal string
`"untrusted"` for a validly signed binary.

This only manifests under `foreach(workers=N)`, so serial collections are correct and the
corruption is invisible unless you diff a parallel run against a serial baseline.

I have a patch and tests ready and would like to contribute them — see "Proposed fix" below.

### The chain

1. `ReaderPool` builds its `ttlcache` with **no TTL**, so every entry keeps the zero
   `expireAt`. `priorityQueue.Less` returns `false` when the left item's `expireAt` is zero, so
   with all items zero the heap never orders, and `SetCacheSizeLimit` eviction takes
   `items[0]` — an arbitrary entry rather than the LRU. `remove()` swaps the *last-pushed*
   item into index 0, so the next victim is preferentially the **most recently created**
   reader, i.e. one likely still in flight.
2. The expiration callback closes that reader from a background goroutine (`go accessor.Close()`).
3. `ReadAt` has already released the mutex, so the close lands on a handle a read is using.
4. `pe.ParseSecurityInfo()` propagates the error, so `authenticode()` skips its entire
   signed-file block and returns its initialisers. Every certificate field is set inside one
   `if err == nil` block, so they all blank together — and `Trusted` is seeded as `"untrusted"`
   at the top of the function rather than null, which is what turns a failed *read* into a
   confident wrong *verdict*.

Serially only one reader is in flight and it is the one just inserted, which is why this is
invisible without `workers=`.

### Affected

Every `readers.NewAccessorReader` caller, not just authenticode: `parse_pe`, `parse_binary`,
`ese`/`srum`, `journald`, `raw_registry`, and the ext4/fat/mscfb/xfs accessors.

For DFIR use this is a correctness issue rather than a performance one: a dropped signer turns
a renamed-binary or provenance hunt into a silent miss, and a flipped `Trusted` turns a signed
Microsoft binary into a false positive in any notebook filtering `Trusted != 'trusted'`.

### Reproduction

Still present on `master` (`cde8b59b5`) — the pattern is visible directly at the end of
`AccessorReader.ReadAt`.

Measured on a **stock v0.75.6 Windows client** (Win10 22H2 VM). Method: run the same
`authenticode()` enrichment over the same file set serially and at `workers=4`, key rows by
path, and diff the certificate fields.

| Distinct paths through the pool | Path-trials | Subject+Issuer lost | Rate |
| --- | --- | --- | --- |
| 100–286 | 5,470 | 4 | 0.073% |
| 410–5,740 | 12,934 | 50 | **0.387%** |

- **Two serial runs were byte-identical across all 5,740 paths** (0 differences), so the
  losses are attributable to concurrency and not to run-to-run instability.
- A different set of files is hit every run (pairwise Jaccard ≈ 0).
- Every loss carried the same signature: `SubjectName` and `IssuerName` null *together*, the
  `Authenticode` object itself non-null, `Trusted` == `"untrusted"`.
- The rate rises with the number of distinct paths pushed through the pool, consistent with
  eviction pressure being the trigger.
- Also observed on a 0.77.1 client (≈1% on a smaller corpus).

Two gotchas for anyone reproducing it:

- **Do not use `%SystemRoot%\System32` as the corpus.** On current Windows it is essentially
  all catalog-signed, so `pe.ParseAuthenticode` never reaches the affected block and the test
  passes vacuously. Use `%ProgramFiles%`.
- **`go test -race` does not catch this.** It is a file-descriptor lifetime race, not a Go
  memory race — every field access is already mutex-guarded, and `-race` reports no
  `DATA RACE` on the unfixed code. Only a value diff against a serial baseline finds it.

### Proposed fix

Hold the mutex for the duration of the read, so `Close()` waits for an in-flight read instead
of pulling the handle out from under it. Reopening *after* a close stays supported — that is
the pool's design and is merely I/O.

```go
-	paged_reader := self.paged_reader
-
-	// Reading from the paged reader may trigger another reader due to
-	// LRU so we release the lock before we do it.
-	self.mu.Unlock()
-
-	return paged_reader.ReadAt(buf, offset)
+	defer self.mu.Unlock()
+
+	return self.paged_reader.ReadAt(buf, offset)
```

This should cost no parallelism: readers are keyed per accessor+path so distinct files still
read concurrently, and go-ntfs' `PagedReader` already serialises every read through its own
mutex — so two callers sharing one `AccessorReader` were serialised before the change too.

I have two regression tests that both fail on the unfixed code (verified by reverting):

- a portable one with no Windows API dependency — on the unfixed pool, 9,370 of 9,600
  concurrent pooled reads failed while the serial baseline stayed clean;
- an end-to-end `authenticode()` one — 35 of 40 embedded-signed binaries corrupted unfixed.

Happy to open a PR with the fix plus the portable test. Please let me know if you would prefer
the eviction-ordering problem addressed instead of (or in addition to) this — note that
`priorityQueue.Less` lives in `Velocidex/ttlcache`, so that would be a separate change. I
think `ReadAt` should be safe regardless, since `Close()` can also arrive from the lifetime
alarm and from another caller's `defer reader.Close()` on the same pooled object.

AlexSta2019:
Fixes #4963, which has the full analysis, the measurements and the reproduction recipe.

### The change

`AccessorReader.ReadAt` grabbed `self.paged_reader` and released the mutex before reading
through it. `Close()` can run concurrently — from the pool's size-limit eviction callback,
from the lifetime alarm, or from another VQL function's `defer reader.Close()` on the same
pooled object — so the read could land on a closed handle and fail with `os.ErrClosed`.

Callers that treat a read error as "not present" rather than "retry" then emit wrong data
with no error and no change in row count. `authenticode()` is the worst case: it returns
`SubjectName`/`IssuerName` null *together* and `Trusted` as the literal `"untrusted"` for a
validly signed binary.

The fix holds the lock for the duration of the read, so `Close()` waits for an in-flight read
instead of pulling the handle out from under it. Reopening *after* a close stays supported —
that is the pool's design and is merely I/O.

**This should cost no parallelism.** Readers are keyed per accessor+path, so distinct files
still read concurrently, and go-ntfs' `PagedReader` already serialises every read through its
own mutex — two callers sharing one `AccessorReader` were serialised before this change too.

### Test

`vql/readers/paged_race_test.go` is portable (no Windows API, no build tags). On the unfixed
pool **9,370 of 9,600 concurrent pooled reads failed**, while the serial baseline stayed
clean; it passes with the fix. Verified by reverting the `paged.go` hunk.

Worth flagging for reviewers: **`go test -race` does not catch this bug.** It is a
file-descriptor lifetime race, not a Go memory race — every field access was already
mutex-guarded, and `-race` reports no `DATA RACE` on the unfixed code. Only a value diff
against a serial baseline finds it, which is why the test is written the way it is.

### Scope note

This deliberately does not touch the eviction-ordering problem described in #4963 (the pool
builds its `ttlcache` with no TTL, so all entries keep the zero `expireAt`, `priorityQueue.Less`
never orders them, and size-limit eviction takes an arbitrary — preferentially
most-recently-created — entry). That lives in `Velocidex/ttlcache` and is a separate change.
`ReadAt` should be safe regardless of which entry eviction picks, since `Close()` also arrives
from the lifetime alarm and from other callers.

### Disclosure

I was not able to compile or run the test against current `master` on the machine I prepared
this on — no Go toolchain available. The hunk applies cleanly to `cde8b59b5` and I verified
statically that every package and symbol the test references exists in `master`
(`utils/tempfile.TempDir`, `constants.SCOPE_ROOT`, `acl_managers.NullACLManager`,
`vql_subsystem.ACL_MANAGER_VAR` / `MakeScope`). The numbers above were measured on a build of
this same change. Please let CI be the judge, and I'm glad to fix anything it turns up.

The CLA has not been signed for this contribution yet — please advise on the process.

### DIFF
--- a/services/debug/profile.go
+++ b/services/debug/profile.go
@@ -7,6 +7,7 @@ import (
 
 	"www.velocidex.com/golang/velociraptor/utils"
 	"www.velocidex.com/golang/vfilter"
+	"www.velocidex.com/golang/vfilter/types"
 )
 
 type ProfileWriter func(
@@ -135,4 +136,33 @@ func GetProfileWriters() (result []ProfileWriterInfo) {
 	defer mu.Unlock()
 
 	return append(result, handlers...)
+}
+
+func GetProfileWriterByeName(name string) *ProfileWriterInfo {
+	mu.Lock()
+	defer mu.Unlock()
+
+	for _, h := range handlers {
+		if h.Name == name {
+			return &h
+		}
+	}
+	return nil
+}
+
+func GetProfile(
+	ctx context.Context, handler ProfileWriter) (res []vfilter.Row) {
+	output_chan := make(chan types.Row)
+	scope := vfilter.NewScope()
+
+	go func() {
+		defer close(output_chan)
+		handler(ctx, scope, output_chan)
+	}()
+
+	for row := range output_chan {
+		res = append(res, row)
+	}
+
+	return res
 }
--- a/vql/readers/paged.go
+++ b/vql/readers/paged.go
@@ -41,6 +41,7 @@ import (
 	ntfs "www.velocidex.com/golang/go-ntfs/parser"
 	"www.velocidex.com/golang/velociraptor/accessors"
 	"www.velocidex.com/golang/velociraptor/utils"
+	"www.velocidex.com/golang/velociraptor/utils/files"
 	vql_subsystem "www.velocidex.com/golang/velociraptor/vql"
 	"www.velocidex.com/golang/vfilter"
 )
@@ -87,6 +88,9 @@ type AccessorReader struct {
 
 	key      string
 	max_size int64
+
+	// Atomic reference count for in flight readers
+	ref int
 
 	reader       accessors.ReadSeekCloser
 	paged_reader *ntfs.PagedReader
@@ -172,6 +176,14 @@ func (self *AccessorReader) Flush() {
 func (self *AccessorReader) Close() error {
 	self.mu.Lock()
 
+	if self.ref > 0 {
+		self.ref--
+		self.mu.Unlock()
+		return nil
+	}
+
+	self.ref = 0
+
 	cancel := self.cancel
 	self.cancel = nil
 
@@ -188,6 +200,8 @@ func (self *AccessorReader) Close() error {
 	}
 
 	if reader != nil {
+		// Track opens and closes
+		files.Remove(self.File.String())
 		reader.Close()
 	}
 
@@ -222,9 +236,21 @@ func (self *AccessorReader) ReadAt(buf []byte, offset int64) (int, error) {
 		paged_reader, err := ntfs.NewPagedReader(
 			utils.MakeReaderAtter(reader), 1024*8, lru_size)
 		if err != nil {
+			reader.Close()
 			self.mu.Unlock()
 			return 0, err
 		}
+
+		// Try to read this buffer from the new reader
+		result, err := paged_reader.ReadAt(buf, offset)
+		if err != nil {
+			reader.Close()
+			self.mu.Unlock()
+			return 0, err
+		}
+
+		// Track opens and closed
+		files.Add(self.File.String())
 
 		// Set an alarm to close the file in the future - this
 		// ensures we don't hold open handles for long running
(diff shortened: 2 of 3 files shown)


scudette:
As the comment said the lock is released before the read because the read may trigger other reads within the same accessor. This leads to a deadlock.

Probably the best option is to implement a reference count to avoid closing the accessor if another user is currently reading.

PR #4974, opened 2026-08-13
Add Azure Monitor (Log Analytics) upload plugin and artifact

Renizmy:
## Summary
- Adds `azure_monitor_upload()`, a new VQL plugin that uploads rows to an Azure Log Analytics workspace via the Logs Ingestion API (Data Collection Rule based), since the existing `adx_upload()` plugin talks to ADX's Kusto ingestion endpoint and cannot be pointed at Log Analytics.
- Adds the `Azure.Monitor.Upload` server-event artifact that watches completed flows and streams matching results to the new plugin.
- Supports service principal, `AZURE_*` environment variables, managed identity, and the Azure SDK default credential chain (the last two gated behind a `sumo` build tag to keep the `azidentity` SDK out of default builds).
- Adds an `Azure Monitor Creds` secret type so credentials/destination can be stored in the secrets service instead of the artifact.

## Test plan
- [x] Unit tests in `vql/server/azure_monitor_test.go` (URL building, retry-after parsing, row marshaling, timestamp fallback edge cases)
- [x] Manual end-to-end run against a real Log Analytics workspace + DCR

### DIFF
--- a/vql/server/azure_monitor_mi.go
+++ b/vql/server/azure_monitor_mi.go
@@ -22,12 +22,10 @@
 /*
 Azure Monitor managed-identity / default-credential auth.
 
-This file is compiled only under the "sumo" build tag because it pulls in the
-azidentity SDK (which is also used by the ADX plugin). It provides the
-azidentity-backed token sources for the managed_identity and default_credential
-auth modes of azure_monitor_upload(). The service-principal path (including
-AZURE_* environment variable credentials) lives in azure_monitor.go and is
-always compiled.
+The azidentity-backed token sources for the managed_identity and
+default_credential auth modes of azure_monitor_upload(). The service-principal
+path (including AZURE_* environment variable credentials) uses
+golang.org/x/oauth2 directly and lives in azure_monitor.go.
 
 Note: azidentity manages its own HTTP stack for token acquisition (IMDS for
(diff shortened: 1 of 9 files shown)


vql/server/azure_monitor_mi.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	"golang.org/x/oauth2"
+	vfilter "www.velocidex.com/golang/vfilter"
+)
+
+func init() {
+	// Managed identity (optionally user-assigned via managed_identity_client_id).
+	azureMonitorMITokenSource = func(
```
scudette: Maybe set these as regular functions?

Probably the entire plugin should not be compiled into the non-sumo build rather than including a broken version of it.

The user can then check for the presence of the plugin using the version() VQL function. It is a good idea to add that check to the artifact (e.g. in a precondition).

Renizmy: Thanks, I’ve updated the implementation

PR #4982, opened 2026-08-18
Wait for final FlowStats write before completion

### LINKED ISSUE #4969: Race condition in final FlowStats write can cause Generic.Client.Info interrogation results to be missed
I noticed that several new clients were missing hostnames. Upon further inspection, the clients appeared to have completed the initial Generic.Client.Info flow requested by the InterrogationService but did not reflect the results in their client info…

### The bug…
When a client connects to the frontend server for the first time, the frontend receives its client_id (the client derives its ID from its public key). The frontend server then publishes the client_id to an internal server queue on the master server (Server.Internal.Enrollment).

The InterrogationService that runs on the master server is subscribed to the queue and sees the new message. It first checks the client record and ensures that LastInterrogateFlowID is empty. It then creates a placeholder client record with the client_id and schedules the Generic.Client.Info artifact (once this returns it should contain more info about the client such as the client’s hostname) by placing it in the client’s task queue. It also marks LastInterrogateFlowID = F.xxx to indicate the interrogation has been started and record the relevant flowID.

As the client runs the artifact it sends results back to the frontend server which queues them up to be written. The Generic.Client.Info artifact has multiple "sources" aka subqueries (BasicInformation, DetailedInfo, LinuxInfo, WindowsInfo, Users) and each result is written to a file under clients/C.xxx/collections/F.xxx/… depending on the subquery. In particular, a write to the shared filestore is queued by the frontend for each batch of results received and a counter (the runner’s Completer) is incremented to indicate a pending write. Several goroutines loop in the background on the frontend server, draining this queue and decrementing the counter after each write to disk.

After all subqueries finish, the client has finished the artifact and it sends a FlowStats message with “FlowComplete: true” to the frontend server to indicate the flow is complete. After the frontend server receives this flow completion status it creates a “summary record” that consists of metadata about the flow and the names of the subqueries that produced results (this is the record that the master server will read and use to determine what subqueries to read in order to populate the client’s info). The frontend server then queues the summary record to be written and sends an update to the master server (once the runner’s Completer has reached 0). (When queuing a write to the datastore, one can specify a callback function that will be called after the write is completed and in this case no callback is provided. If no callback function is provided, the runner’s Completer is not incremented.)

The master server then gets the message from the frontend that the flow is complete and goes to check stats.json.db to see which sources produced results.
**However a race condition exists here where if the write to the datastore for the summary results is not yet written to disk, the master server will check clients/C.xxx/collections/F.xxx/stats.json.db and see that it does not list that some of the subqueries produced results. As a result, the master server will not look at some of the subqueries’ results and will not populate any client info (for example, if stats.json.db does not indicate that the source BasicInformation produced results, the master server will not look at clients/C.xxx/artifacts/Generic.Client.Info/F.xxx/BasicInformation.json that contains the client’s hostname).**

### Proposed solution…
If the frontend server were to instead include self.completer.GetCompletionFunc() as the callback function, the runner’s Completer would be incremented to indicate a write is still in flight and the message to the server that the flow has been completed would not be sent until the write has occurred.

_Validated by adding a regression test that blocks the final stats write and confirms System.Flow.Completion is not emitted until the write completes, and by confirming a localhost client/server enrollment successfully populated Generic.Client.Info/BasicInformation and hostname after the patch._

_flows/client_flow_runner.go_ 
Old code:
```
err = launcher_service.Storage().WriteFlowStats(ctx, self.config_obj, stats, utils.BackgroundWriter)
if err != nil {
	return err
}
```
New code:
```
 launcher_service, err := services.GetLauncher(self.config_obj)
 if err != nil {
 	return nil
 }
// if doing completion write, register with completer so flow finished msg is not sent until record written to disk
completion := utils.BackgroundWriter
if msg.FlowComplete {
completion = self.completer.GetCompletionFunc()
}
err = launcher_service.Storage().WriteFlowStats(ctx, self.config_obj, stats, completion)

if err != nil {
 	return err
 }
```

aglenn-js:
Fixes [issue #4969](https://github.com/Velocidex/velociraptor/issues/4969) ("Race condition in final FlowStats write can cause Generic.Client.Info interrogation results to be missed")

- Wait for the final FlowStats datastore write before publishing System.Flow.Completion event
- Add test that blocks the final write and verifies completion event is held until final write is complete

### DIFF
--- a/flows/client_flow_runner.go
+++ b/flows/client_flow_runner.go
@@ -612,8 +612,12 @@ func (self *ClientFlowRunner) FlowStats(
 		return nil
 	}
 
-	err = launcher_service.Storage().WriteFlowStats(ctx, self.config_obj,
-		stats, utils.BackgroundWriter)
+	// if doing completion write, register with completer so flow finished msg is not sent until record written to disk
+	completion := utils.BackgroundWriter
+	if msg.FlowComplete {
+		completion = self.completer.GetCompletionFunc()
+	}
+	err = launcher_service.Storage().WriteFlowStats(ctx, self.config_obj, stats, completion)
 	if err != nil {
 		return err
 	}
--- a/flows/client_flow_runner_test.go
+++ b/flows/client_flow_runner_test.go
@@ -14,6 +14,7 @@ import (
 	"github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
+	"google.golang.org/protobuf/proto"
 	actions_proto "www.velocidex.com/golang/velociraptor/actions/proto"
 	"www.velocidex.com/golang/velociraptor/api"
 	api_proto "www.velocidex.com/golang/velociraptor/api/proto"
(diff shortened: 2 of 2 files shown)


aglenn-js:
Hi! I see there is a CLA to sign but I am making this contribution on behalf of my employer and I am not the authorized signer. I see that the CLA.md mentions a separate Corporate CLA with Velocidex. Can someone please provide the Corporate CLA process?

scudette:
Sorry this is the only CLA we have.

aglenn-js:
Thanks for clarifying. It looks like the automated check ensures that the PR contributor has signed the CLA, but is it possible for the authorized signer at my company to sign the CLA on my behalf? If so, should they use the same PR link while signed into their own GitHub account, and would it be possible to then mark aglenn-js as covered so the automated check passes?

scudette:
The GitHub CLA checker is just a tool to make it easier - you can just send an email to support@velocidex.com if that it easier. It doesn't need to be too complicated, just mention the cla and that you (or your company) agree to it.

aglenn-js:
Sweet, I just sent over an email. Please let me know if anything else is needed from me

PR #4986, opened 2026-08-22
VQL language server: smarter completions and a self-healing gRPC proxy

predictiple:
## Completions
- Argument names now complete inside unclosed calls (`pslist(`) and behind trigger dots (`parse_lines(.`), each carrying a `textEdit` so the editor replaces any typed prefix cleanly.
- `(` triggers argument completion for queries that don't parse yet; signature help stays silent there so the popups never overlap.
- Positional filtering keeps functions out of `FROM` positions; partial typing (`parse_lines(fi`) filters args by name instead of matching them all equally.

## gRPC proxy resilience
- The proxy caches open documents and replays them if the backend loses state, so frontend restarts no longer silently break completions until the file is reopened.
- A watchdog pings the backend every 10s; a stuck connection is hot-swapped for a fresh dial rather than black-holing editor requests.
- Forwarded calls carry a 30s deadline, and the previously missing shutdown/exit lifecycle handlers mean the language server stops cleanly instead of being hard-killed after a timeout.

All verified end-to-end with scripted VS Code sessions (keystroke injection) plus wire-level regression checks.


services/lsp/api.go
```diff
@@ -14,6 +14,9 @@ import (
 	vql_subsystem "www.velocidex.com/golang/velociraptor/vql"
 	"www.velocidex.com/golang/vfilter"
 	"www.velocidex.com/golang/vfilter/types"
+
```
scudette: breaks the build

predictiple: fixed in https://github.com/Velocidex/velociraptor/pull/4986/changes/251b6c682400fb3091f9e1cb6406879e2a0983a3

PR #5003, opened 2026-09-08
Bugfix: transform hostname in pool client compressed responses

predictiple:
The pool client's hostname transform silently no-oped for newer clients because the launcher compresses VQL responses by default (FlowRequest_ZLIB). The payload lives in CompressedJsonResponse while JSONLResponse is empty, so the transform parsed an empty string and returned the response unchanged.

Decompress the payload before transforming and re-compress the result, preserving the original compression state.

History:

- The bug has existed since July 2025, when an unrelated filestore change (#4361) made compressed flow requests the default: the pool client's hostname transform, written in 2023 when responses were uncompressed, silently stopped working because it only read `JSONLResponse` while the payload now lived in `CompressedJsonResponse`. The failure was completely invisible - parsing the empty string returns no error, so the transform passed every response through unchanged without a single log line - and the symptom (all virtual clients showing the same hostname) was indistinguishable from expected behavior, since the pool client's entire premise is many simulated clients running on one real host. Combined with the fact that the pool client is a load-testing tool whose operators watch server metrics rather than client hostnames, the `-<id>` suffix was a purely cosmetic nicety whose absence looked normal, and the only alternative hostname path (the periodic `GetClientInfo()` update) was disabled by default, so nothing ever surfaced the discrepancy.

It surfaced now because I'm experimenting with giving pool clients unique hostnames, so that the feature can also be useful in training and development environments :sunglasses:

### DIFF
--- a/executor/pool.go
+++ b/executor/pool.go
@@ -128,8 +128,7 @@ func maybeTransformResponse(response *actions_proto.VQLResponse, id int) *action
 			// case the JSONL payload is stored in
 			// CompressedJsonResponse and JSONLResponse is empty.
 			jsonl := response.JSONLResponse
-			compressed := len(response.CompressedJsonResponse) > 0
-			if jsonl == "" && compressed {
+			if jsonl == "" && len(response.CompressedJsonResponse) > 0 {
 				decompressed, err := utils.Uncompress(
 					context.Background(), response.CompressedJsonResponse)
 				if err != nil {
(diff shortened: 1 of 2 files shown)


executor/pool.go
```diff
@@ -144,7 +158,20 @@ func maybeTransformResponse(response *actions_proto.VQLResponse, id int) *action
 				return response
 			}
 			result := proto.Clone(response).(*actions_proto.VQLResponse)
-			result.JSONLResponse = string(new_rows)
+
+			// Preserve the compression state of the original
```
scudette: we probably don't really need to recompress it back - the server code will fall back to handle uncompressed data anyway so we just need to clear the compressed field and pass it an uncompressed

predictiple: You're right. I've added a new commit to skip that.

PR #5007, opened 2026-09-09
Add unique emulated identities to pool clients

predictiple:
While the pool client is designed for _server_ load testing, it has shown itself to be useful for testing/learning/development scenarios too (for example, testing hunt behaviours).

This PR preserves the load-testing aspects but adds some features that make the pool client more human-friendly and therefore more useful for lab and learning environments.

Each pool client now presents a unique hostname and FQDN (e.g. "quantum-phoenix.example.com") to the server instead of the real host's hostname suffixed with the client id. The FQDN reuses the domain of the host running the pool client so the emulated clients appear to belong to the same network. Hostnames are adjective-noun combinations (e.g. "quantum-phoenix") - a numeric suffix is only appended when the number of clients is almost ten thousand. Name collisions are possible if running pool clients simultaneously on multiple hosts connected to the same server, but the server is designed to expect name collisions in the real world, so it's not an issue.

Identities are generated upfront for all clients and applied to the `BasicInformation` and `DetailedInfo` sources of the `Generic.Client.Info` interrogation, as well as the periodic client info messages.

Writebacks now default to in-memory only (`memory:// scheme`) so no writeback files are written to disk and each client's identity is ephemeral.  This makes it easy to spin up clients -> run some tests -> stop the clients -> delete the data -> repeat, without needing to go clean up hundreds of writeback files each time. This also makes it easy to create massive amounts of hunt data in batches (create hunt -> add 10,000 clients -> hunt collections run -> stop clients -> add 10,000 new clients -> hunt collections run -> stop clients -> repeat) for server-side query performance testing.

The writeback location is now configured for all platforms - previously Darwin was left unset and fell back to the config file's path, causing all clients on macOS to share a single writeback file. The root config loader no longer requires a writeback file, but it is still available if you need persistent pool client identities.

Screenshots:

pool_client

pool_client2

### DIFF
--- a/bin/pool.go
+++ b/bin/pool.go
@@ -46,8 +46,12 @@ var (
 		"number", "Total number of clients to run.").Int()
 
 	pool_client_writeback_dir = pool_client_command.Flag(
-		"writeback_dir", "The directory to store all writebacks. "+
-			"If not specified, in-memory writebacks are used.").ExistingDir()
+		"writeback_dir", "The directory to store all writebacks.").Default(".").
+		ExistingDir()
+
+	pool_client_no_writeback = pool_client_command.Flag(
+		"no_writeback", "Use in-memory writebacks so client identities are ephemeral.").
+		Bool()
 
 	pool_client_concurrency = pool_client_command.Flag(
 		"concurrency", "How many real queries to run.").Default("10").Int()
@@ -74,6 +78,10 @@ func (self *counter) Inc() {
 
 func doPoolClient() error {
 	logging.DisableLogging()
+
+	if *pool_client_no_writeback && *pool_client_writeback_dir != "." {
+		return fmt.Errorf("--no_writeback and --writeback_dir are mutually exclusive")
+	}
 
 	number_of_clients := *pool_client_number
 	if number_of_clients <= 0 {
@@ -137,17 +145,18 @@ func doPoolClient() error {
 
 			client_config.Client.DisableCheckpoints = true
 
-			// Use an in-memory writeback so each client's identity is
-			// ephemeral and no files are written to disk. If the user
-			// specified a writeback directory, write the writebacks
-			// there instead so identities persist between runs.
-			if *pool_client_writeback_dir != "" {
+			// By default write the writebacks to the writeback
+			// directory so client identities persist between runs.
+			// If --no_writeback is specified, use an in-memory
+			// writeback so each client's identity is ephemeral and no
+			// files are written to disk.
+			if *pool_client_no_writeback {
+				client_config.Client.WritebackLinux = fmt.Sprintf(
+					"memory://pool_client.yaml.%d", i)
+			} else {
 				client_config.Client.WritebackLinux = path.Join(
 					*pool_client_writeback_dir,
 					fmt.Sprintf("pool_client.yaml.%d", i))
-			} else {
-				client_config.Client.WritebackLinux = fmt.Sprintf(
-					"memory://pool_client.yaml.%d", i)
 			}
 			client_config.Client.WritebackWindows = client_config.Client.WritebackLinux
 			client_config.Client.WritebackDarwin = client_config.Client.WritebackLinux
@@ -167,9 +176,9 @@ func doPoolClient() error {
 			}
 			client_config.Client.Concurrency = uint64(*pool_client_concurrency)
 
-			// Register the in-memory writeback with the writeback
-			// service. This generates a fresh private key and client
-			// id for this client.
+			// Register the writeback with the writeback service. This
+			// generates a fresh private key and client id for this
+			// client if no writeback exists yet.
 			writeback_service := writeback.GetWritebackService()
 			_ = writeback_service.LoadWriteback(client_config)
 
--- a/bin/pool.go
+++ b/bin/pool.go
@@ -23,6 +23,7 @@ import (
 	"path"
 	"sync"
 
+	"github.com/Showmax/go-fqdn"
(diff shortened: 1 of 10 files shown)


scudette:
The reason to preserve the write back files from run to run is for efficiency - the initial client enrollment is quite expensive (a lot of RSA operations) so on a single host this takes quite a while. Ultimately the write backs only store private/public key pairs (which influence the client id) so it is actually completely safe to reuse them.

predictiple:
> The reason to preserve the write back files from run to run is for efficiency - the initial client enrollment is quite expensive (a lot of RSA operations) so on a single host this takes quite a while. Ultimately the write backs only store private/public key pairs (which influence the client id) so it is actually completely safe to reuse them.

Ok but I hate having to go clean them up every time I want a fresh test (after deleting the clients on the server). The writeback is just optional now. In most situations where you'd use pool client it's likely to be in the dozens to 100s range. Rarely 10,000 or a million.

Also for new users (like someone who's curious about Velociraptor and wants to set up a home lab to play around) they don't need to break their brains trying to figure out where the writeback actually lives (`C:\Temp` or `C:\Windows\Temp` or `%USERPROFILE%\AppData\Local\Temp` or `C:\Program Files\Velociraptor\tools`?). It's intuitive that when you run a new pool client session you get X number of new pool clients.

scudette:
Why would you even need to clean the write back files up?

This is actually important because the first run creating the keys would take so long that it would not be possible to emulate a large number of clients enrolling upon first deployment. So for testing this I would usually create a few thousand keys first, then clear the server data store and then run it again. The second run would be very fast because the pool client would be already initialized.

predictiple:
As I mentioned in the original comment: a scenario where you have limited hardware but want to have far more clients represented in the datastore than your hardware will allow to run simultaneously. Ephemeral writebacks allows you to batch-create the hunt data.

Another scenario where the writeback files are any annoyance:
- I start a pool client to test a hunt. All the collections happen.
- I make adjustments to the hunt and want to restest with new clients. So I delete the existing clients because I don't want to have to figure out which are old clients vs new clients.
- Restart the pool client. Clients come back online but they've already participated in the hunt so they don't do it again.
- Deleting the writeback files provides the desired "fresh start".

I see your point for performance testing. In your scenario that's still exactly possible but you just need to use `--writeback_dir` if you want that behaviour. Ephemeral writebacks serve the non-collection-performance-testing scenarios better IMHO, but also allows generating massive amounts of data for testing other things on the server which would be painful to achieve currently (you could write a script that starts pool_client, waits, stops it, deletes writebacks, starts again, etc.).

predictiple:
Some realworld data:

## 100 clients

| Writeback                | Start rate |   N |     Time | ms/client |
|:-------------------------|:----------:|----:|---------:|----------:|
| memory:// (ephemeral)    |    20/s    | 100 | 10,350ms |       103 |
| --writeback_dir (cached) |    20/s    | 100 |  5,347ms |        53 |

## 10,000 clients

| Writeback                | Start rate |      N |                  Time | ms/client |
|:-------------------------|:----------:|-------:|----------------------:|----------:|
| memory:// (ephemeral)    |  unlimited | 10,000 | 1,156,881ms (~19.3 min) |       116 |
| memory:// (ephemeral)    |    20/s    | 10,000 | 1,266,606ms (~21.1 min) |       127 |
| --writeback_dir (cached) |    20/s    | 10,000 |   500,456ms (~8.3 min) |        50 |
| --writeback_dir (cached) |  unlimited | 10,000 |    75,371ms (~1.3 min) |       7.5 |

- Key generation costs ~108ms per client and is serialized by a global mutex
  in LoadWriteback (services/writeback/writeback.go:77).
- Cached client setup is only ~7.5ms per client.
- At the default 20/s throttle the cached ramp is throttle-limited
  (10,000 / 20 = 8.3 min), but the ephemeral ramp is key-gen-limited
  (~9.3 clients/s effective) because key gen (108ms) exceeds the throttle
  interval (50ms).

For large scale testing you've got a significant startup time no matter what you do. It depends what you're testing but most of the time you'd have to account for the startup time in your test results anyway (it's never close to zero).

If that was me I'd go have a 2nd :smoking: :sunglasses:

predictiple:
It's also worth noting that the serialized keygen is a design choice rather than an inherent constraint. We could parallelize that (in a separate change). But maybe there's some other reasoning for it being the way that it is?

scudette:
It depends what you are testing for sure - there are two phases to the deployment:
1. Enrol all the clients 
2. Run a particular hunt

As you pointed out enrolment is rate limited so there is a significant start up time for first enrolment. However, this is not usually what is being tested - so it makes sense to skip that and go right to the hunt part. If the write back is ephemeral it means every single start is expensive (even if you don't wipe the data store). This leads to a lot of extra clients in the data store which requires constant cleaning of it as well.

It just makes sense to keep the write backs most of the time because restarting the pool client is a very frequent operation, while testing client key gen is very rare. Maybe keep it behind a flag?

scudette:
From a usability perspective, ephemeral clients make it very hard to work with the same clients after a pool client restart.

For example if you look at the exercise we did in the training module around stacking this will not be possible to do because it requires restarting of the pool client, but we need to keep using the same client IDs to demonstrate change between one collection and the next.

predictiple:
Fair enough. I'll probably use ephemeral most of the time but there are definitely situations where it could cause problems, like your training example.

I've switched the default in my last commit. I added the flag `--no_writeback` which is technically not correct but more concise/elegant than `--no_writeback_files`. Hope that's ok?

PR #5016, opened 2026-09-15
Fix Windows.Registry.EnableUnsafeClientMailRules undefined OSPath

### LINKED ISSUE #3948: Fix Windows.Registry.EnableUnsafeClientMailRules
Looks like there is an error in Windows.Registry.EnableUnsafeClientMailRules - We should look at adding this to RegistryHunter
Possible recursive bug during a refactor https://github.com/Velocidex/velociraptor/commits/master/artifacts/definitions/Windows/Registry/EnableUnsafeClientMailRules.yaml

cristianchiriac:
Fixes #3948

## Problem
`Windows.Registry.EnableUnsafeClientMailRules` never returns results. Every
user profile logs `Symbol OSPath not found` and the collection ends with an error.

## Root cause
The `foreach` row only contains `Username` and `NTUser`, but `read_reg_key()`
was rooted at `pathspec(DelegatePath=OSPath)`. `OSPath` is not in scope there.
`NTUser` already holds the hive path, because the single-column subquery
collapses to its value.

## Fix
Use `DelegatePath=NTUser`.

## Testing
With the v0.77.2 Windows release binary, run as Administrator:
- Built-in artifact: `Symbol OSPath not found` for every profile.
- Patched artifact (`--definitions`) on live profiles: no errors.
- Same query shape on a test hive with
  `Software\Microsoft\Office\16.0\Outlook\Security\EnableUnsafeClientMailRules = 1`:
  the patched version returns the row. The old expression fails with the same error.

The artifact had no golden test. I'm happy to add one with a hive fixture if you'd like.


scudette:
this probably needs to be absorbed into the registry hunter eventually since it is just a registry based artifact.

Tatamis:
Agreed — folding this into the registry hunter makes sense as the end state. This PR is deliberately minimal in the meantime: a one-character change of `DelegatePath=OSPath` to `DelegatePath=NTUser`, so the existing artifact stops failing with `Symbol OSPath not found` for every profile. Happy for it to be superseded by the hunter work whenever you get to it, or to close it if you'd rather fix it there in one go.

One note on CI: the failing Windows Test job is a nil-pointer panic in `http_comms`, which this artifact-only change can't reach.

